PR #16, opened 2013-05-16
Feature/js syntax

indutny:
/cc @veged


indutny:
Aaargh, wrong base...

PR #22, opened 2013-05-22
i-bem: Fixed wrong regexp matching issue in `_buildModValRE`

corpix:
Port [#363](https://github.com/bem/bem-bl/pull/363) from `bem-bl`


dfilatov:
спасибо!

PR #19, opened 2013-06-06
Add block `page` for html/head/body stuff

veged:


common.blocks/page/__css/page__css.bemhtml
```diff
@@ -0,0 +1,41 @@
...
+                }));
+            } else {
+                var hideRule = !ie ?
+                        ['gt IE 9', '<!-->', '<!--'] :
+                        ie === '!IE' ?
+                            [ie, '<!-->', '<!--'] :
+                            [ie, '', ''];
+                apply(
+                    '',
+                    { 'ctx._ieCommented': true },
```
narqo, line 24: Where could I read about things that happens here?

common.blocks/page/page.bemhtml
```diff
@@ -0,0 +1,67 @@
+block('page')(
+
+    mode('doctype')(this.ctx.doctype || '<!DOCTYPE html>'),
+
+    mode('xUACompatible')(
```
narqo, line 5: Still don't thing that this is a "core-library"'s feature

common.blocks/page/page.bemhtml
```diff
@@ -0,0 +1,67 @@
...
+    }),
+
+    tag()('body'),
+
+    mix()([{ elem: 'body' }]),
+
+    elem('root')(
+        bem()(false),
+        tag()('html'),
+        cls()('i-ua_js_no i-ua_css_standard')
```
narqo, line 47: Should we really need this?

common.blocks/page/page.deps.js
```diff
@@ -0,0 +1,13 @@
+({
+    mustDeps: [
+        {
+            block: 'i-bem',
+            elems: [
+                'html',
+                { elem: 'dom', mods: { init: 'auto' } },
+            ]
+        },
+        { block: 'i-ua' }
```
narqo, line 10: ???

common.blocks/page/page.examples/10-page_simple.bemjson.js
```diff
@@ -0,0 +1,59 @@
+([
+    {
+        block: 'page',
+        mods: { exp: 'ololo' },
+        js: { a: 'b' },
+        title: 'заголовок',
+        head: [
+            {
+                elem: 'css',
+                url: '_simple.css'
```
narqo, line 10: How does this works for bundle `10-page_simple`?

common.blocks/page/page.examples/10-page_simple.bemjson.js
```diff
@@ -0,0 +1,59 @@
...
+            },
+            {
+                elem: 'css',
+                url: '_simple.ie.css',
+                ie: true
+            },
+            { block: 'i-jquery', elem: 'core' },
```
narqo, line 17: `i-jquery`?

narqo:
1) Examples' levels should have `.bem/level.js`-config, which is "simple"-based, I believe.
2) Examples names should not contains "_"-mark (like `30-page_with-js`), until we want to create custom naming-schemes for there's levels...

common.blocks/page/__css/page__css.bemhtml
veged, line 24: What did you mean? About guargind for omit recursion?

common.blocks/page/page.bemhtml
veged, line 47: what can be alternative?

common.blocks/page/page.deps.js
veged, line 10: fix all `i-ua` occurrences

common.blocks/page/page.examples/10-page_simple.bemjson.js
veged, line 17: fixed

veged:
fix examples

common.blocks/page/page.bemhtml
narqo, line 47: @veged `mix`?

common.blocks/page/__css/page__css.bemhtml
narqo, line 24: The syntax is **horrible** – it really looks like a syntax mistake

common.blocks/page/page.bemhtml
veged, line 47: it's a little bit hacky (because we work with it later as raw `className`), so I decide to use `cls`

common.blocks/page/__css/page__css.bemhtml
veged, line 24: yep :-/ I already create https://github.com/bem/bem-xjst/issues/2

common.blocks/page/page.bemhtml
tadatuta, line 47: btw, why don't we want to make it a real mix to have possibility for `findBlockOutside('i-ua').someMethod()`, etc?

veged, line 47: @tadatuta couse there is no real `ua` block with js here -- it's only in `touch.blocks` (where it's fair js-block mixed to `page`)

tadatuta, line 47: My question was about why not to implement it as such block? E.g. we could utilize it to check svg support from inside js.

veged, line 47: we just try to keep it as light-weight as possible all the way -- but if we find enough benefits from using i-bem.js-based block we can do that

btw, I already port [ua__svg](https://github.com/bem/bem-core/blob/dev/common.blocks/ua/__svg/ua__svg.bemhtml)

PR #27, opened 2013-06-11
Move base bemhtml-templates from `i-bem__html` element to block `i-bem` itself

veged:


veged:
we also must fix all deps from `{ block: 'i-bem', elems: 'html' }`

PR #55, opened 2013-06-11
BEMHTML: Add debug files paths to build result

veged:


veged:
cc @narqo @indutny

narqo:
Guess it could be merged with safe

PR #58, opened 2013-06-12
Port i18n from `bem-bl`

arikon:


veged:
it's ok to store libs in elem folder

veged:
do you test it on build `bem-core/common.bundles` itself?

arikon:
@veged No, I haven't tested it yet

dfilatov:
close in favor #239

PR #60, opened 2013-06-12
i18n version of browser.js+bemhtml tech

veged:
Depend on #58 and #53


veged:
cc @arikon @narqo @dfilatov -- we need to rethink full stack of `i18n.*` techs in case of combining (not inheriting) techs

right now it's very hard to write `i18n.browser.js+bemhtml.js` without copy-paste from `i18n.js.js`

dfilatov:
close in favor #239

PR #202, opened 2013-08-07
Issues/201@v1. Translate MIGRATION.md into English

varya:


varya:
Could you check that all the tech terms are OK?

PR #233, opened 2013-08-22
Use only v2 techs during build

SevInf:
Re-configure project ot use basic v2 techs from bem-tools and bem-pr. Depends on narqo/bem-pr#45


dfilatov:
:thumbsup:

PR #234, opened 2013-08-22
i-bem: introduce `this.reapply(context)`

indutny:
see https://github.com/bem/bem-xjst/issues/4

### DIFF
--- a/common.blocks/i-bem/i-bem.bemhtml
+++ b/common.blocks/i-bem/i-bem.bemhtml
@@ -255,6 +255,9 @@ exports.apply = BEMContext.apply = function _apply(context) {
     ctx.apply();
     return ctx._buf.join('');
 };
+
+BEMContext.prototype.reapply = BEMContext.apply;
+
 }); // oninit
 
 match(this._mode === '')(


narqo:
:+1: 

@veged ?

veged:
I miss the point — why it calls REapply? Tme meaning of applying one more time is more suitable for applyNext ;-) Maybe applyNew or newApply?

indutny:
I feel like `reapply` is a better naming for it, as it has almost nothing in common with `applyNext` and `applyCtx`. Because its outside of `xjst`'s subject, and basically the same as calling `exports.apply` whatever it is.

narqo:
I like `reaplly` if anyone cares, but I think that we will get a lot of headache with all of this applies: `apply`, `applyNext`, `applyCtx`, `this.apply` and `this.reapply` for now. 'Cause no one cares where it came from, from xjst or plain-js or somewhere else.

veged:
2 against 1 for `reapply` -- so I humble, let's merge it

@indutny can you rebase and merge?

indutny:
Thanks, landed in 7584dfc.

PR #255, opened 2013-09-15
add pointerdown event for ie < 9 (#253)

### LINKED ISSUE #253: ie < 9 don't work pointerdown event
In ie < 9  don't work pointerdown event, uses in button block from bem-components

fi11:
fixes #253


dfilatov:
closed in favor #283

PR #256, opened 2013-09-15
add pointerup event for ie < 9 (#254)

### LINKED ISSUE #254: ie < 9 don't work pointerup event
In ie < 9 don't work pointerup event, uses in button block from bem-components

fi11:
fixes #254


dfilatov:
closed in favor #283

PR #323, opened 2013-12-03
Up to bem-pr v0.5-beta1

narqo:
\cc @dfilatov check please, if everything is fine, i'll release `bem-pr 0.5.0` and update in he and in bem-components@v2


dfilatov:
it looks good for me

PR #445, opened 2014-03-12
Update path to CDN. Closes #444

### LINKED ISSUE #444: Update path to CDN
`yandex.st` -> `yastatic.net`

tadatuta:


narqo:
@tadatuta 

Don't you think that it is a little bit strange to change CDN's host in open projects until there would be some official news about `yandex.st` at api.yandex.ru?

tadatuta:
@narqo both hosts work fine for some time already and according to our internal blog there are few services who switched. so i don't think it's something secret.

narqo:
@tadatuta the question was not about secrecy but about clearness of the fact that CDN host was changed for "community" ;)

tadatuta:
@narqo if it won't be announced officially until next release we can add some basic explanation in changelog.

narqo:
Sound like a plan =)

PR #446, opened 2014-03-13
Use the same object for this.mods and this.ctx.mods within default mode

### LINKED ISSUE #441: i-bem.bemhtml: use the same object for `this.mods` and `this.ctx.mods`
use

```
mods : vBlock? this.ctx.mods || (this.ctx.mods = {}) : this.mods
```

in `i-bem.bemhtml`

This will allow us to change `mods` object within `default` mod so this changes will reflect node's HTML-classes:

```
block('b1')(
  default()(function() {
    this.mods.a = 'b';
    applyNext();
  }),
  mod('a', 'b')(function() { return 'beep' })
)

// <div class="b1 b1_a_b">beep</div>
```

narqo:
closes #441 

Code style fixed as well

\cc @veged @indutny


narqo:
Tests added for this case

narqo:
Should port this to `i-bem.bemtree` file too

PR #457, opened 2014-03-20
i-bem: events for modifiers changes should be emitted after calling of onSetMod handlers (close #454)

### LINKED ISSUE #454: i-bem: events for modifiers changes should be emitted after calling of onSetMod handlers

dfilatov:
cc @narqo @veged


narqo:
Can we write tests to illustrate the problem? ;)

dfilatov:
I've add test

narqo:
lg

PR #482, opened 2014-04-01
fix issue #481 and add spec for it. Closes #481

### LINKED ISSUE #481: i-bem: declare element by block
``` javascript
var Block = BEMDOM.decl('block'),
    Elem = Block.decl({ elem: 'elem' });

Elem.getName() // => 'undefined__elem', but should be 'block__elem'
```

aristov:


aristov:
https://github.com/bem/bem-core/issues/481

PR #485, opened 2014-04-03
i-bem: export `BEMContext` to oninit's context

### LINKED ISSUE #483: i-bem.bemhtml: Add ability to modify context prototype
We need possibility to modify `BEMContext` in user templates `oninit`. Looks like we can just add `exports.BEMContext = BEMContext` to https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemhtml#L257.

indutny:
fix #483

### DIFF
--- a/common.blocks/i-bem/i-bem.bemhtml
+++ b/common.blocks/i-bem/i-bem.bemhtml
@@ -1,5 +1,5 @@
 /*global oninit */
-oninit(function(exports) {
+oninit(function(exports, context) {
 
 var BEM_ = {},
     toString = Object.prototype.toString,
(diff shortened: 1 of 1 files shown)


common.blocks/i-bem/i-bem.bemhtml
```diff
@@ -1,5 +1,5 @@
 /*global oninit */
-oninit(function(exports) {
+oninit(function(exports, context) {
```
veged, line 2: does second `context` parameter was here all the time?

indutny, line 2: For quite awhile: https://github.com/veged/xjst/commit/358a1d46 :D

narqo:
@veged ping!

veged:
@indutny remind me, does it ported to bem-bl?

PR #490, opened 2014-04-08
i-bem__dom_init_auto: Make async call because of rendering performance (close #486)

### LINKED ISSUE #486: i-bem__dom_init_auto: Make async call because of rendering performance
Client-side performance research in of our `web3` (in compare to `web1`) revealed interesting issues. In some browsers (at least webkit-based) binding of some heavy js-functionality to dom-ready event cause decelerations of rendering process. We compare `$(function() { init() })` with `$(function() { setTimeout(function() { init() }, 0)` and catch ~30% difference in rendering time.

Need port to `bem-bl`.

dfilatov:
cc @veged

### DIFF
--- a/common.blocks/i-bem/__dom/_init/i-bem__dom_init_auto.js
+++ b/common.blocks/i-bem/__dom/_init/i-bem__dom_init_auto.js
@@ -2,10 +2,12 @@
  * Auto initialization on DOM ready
  */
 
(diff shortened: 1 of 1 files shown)


narqo:
@veged do we want this to be a part of "2.2.0"? I'm going to start release preparation today

veged:
it's backward compatible and useful — I merge it

PR #493, opened 2014-04-12
Keyboard helpers (closes #431)

### LINKED ISSUE #431: Keyboard helpers
Think about separate module with common key codes, like enter, arrow keys, etc

narqo:
\cc @dfilatov 

Should I write specs for this? ;)

### DIFF
--- a/common.blocks/keyboard/__codes/keyboard__codes.js
+++ b/common.blocks/keyboard/__codes/keyboard__codes.js
@@ -1,0 +1,25 @@
+/**
+ * @module keyboard__codes
+ */
+modules.define('keyboard__codes', function(provide) {
+
+provide(/** @exports */{
+    BACKSPACE : 8,
+    TAB : 9,
+    ENTER : 13,
+    CAPS_LOCK : 20,
+    ESC : 27,
+    SPACE : 32,
+    PAGE_UP : 33,
+    PAGE_DOWN : 34,
+    END : 35,
+    HOME : 36,
+    LEFT : 37,
+    UP : 38,
+    RIGHT : 39,
+    DOWN : 40,
+    INSERT : 41,
+    DELETE : 42
+});
+
+});


dfilatov:
I think `keyboard` is too general name for module that provides key codes. Perhaps, it would be better if we have `keyboard__keys` module. It also ables us to remove `KEY_` prefix from constants.

``` js
modules.define('keyboard__keys', function(provide) {

provide(/** @exports */{
    BACKSPACE : 8,
    TAB : 9,
    ...
});

});
```

narqo:
Ok, I thought about `keyboard__codes` but decided that it will be too complicated.

The other question is, should we have API for this kind of tasks (we have a huge amount of this crap internally, mainly for a11y staff to work):

``` js
this.bindTo('action', 'keyup', function(e) {
  if(e.shiftKey || e.ctrlKey || e.altKey || e.metaKey) {
    return;
  }
  // ...
})
```

dfilatov:
keyboard__codes looks better than keyboard__keys :)

common.blocks/keyboard/__keys/keyboard__keys.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * @module keyboard__keys
+ */
+modules.define('keyboard__keys', function(provide) {
+
+provide(/** @exports */{
+    BACKSPACE : 8,
```
persidskiy: Why do you use uppercase for object keys? Maybe using lowercase will be quite?

narqo: It's good practice to use uppercase for constants. Actually we always write our code in this way: https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/__internal/i-bem__internal.vanilla.js#L57-L58

narqo:
> keyboard__codes looks better than keyboard__keys :)

For me the main question is how to call required variable for define/require's callback? 

``` js
modules.require(
  ['keyboard', 'keyboard__keys', 'keyboard__codes'],
  function(keyboard, keys, kbCodes) {

keyboard.ENTER; 
keys.ENTER; 
kbCodes.ENTER;

});
```

dfilatov:
`keyCodes.ENTER`?

narqo:
> `keyCodes.ENTER?`

Why not :) :up:

common.blocks/keyboard/__keys/keyboard__keys.js
persidskiy: Ok

dfilatov:
:shipit:

PR #495, opened 2014-04-13
Add Array#find shim to ecma__array

### LINKED ISSUE #494: Implement some of ES6 features as polyfills
As we build a future-aimed library with out strict deadlines, why not to think about implementing some things from, well,.. ~~"future"~~, ES6 standard :)

For things like [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) ([polyfill](https://github.com/paulmillr/Array.prototype.find)) and [`Array.prototype.findIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) ([polyfill](https://github.com/paulmillr/Array.prototype.findIndex)) we commonly use `Array.prototype.some` or plain `while`-loop, which "could look not very clear in some cases" (© @dfilatov).

Firefox supports both of them starting from version 25. I believe Chrome would support them natively to autumn.

narqo:
closes #494 

@dfilatov what would you say?

### DIFF
--- a/common.blocks/ecma/__array/ecma__array.spec.js
+++ b/common.blocks/ecma/__array/ecma__array.spec.js
@@ -1,0 +1,56 @@
+modules.define('spec', ['chai', 'sinon'], function(provide, chai, sinon) {
+
+var should = chai.should();
+
+describe('ecma__array', function() {
+    describe('find', function() {
+        it('should return correct result', function() {
+            var arr = [1, 2];
+            arr.find(function(val) { return val === 2; }).should.be.equal(2);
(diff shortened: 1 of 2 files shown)


dfilatov:
How it will be used? I need to write depependencies in each module where i use it?

narqo:
I think yes. At least until we find a good solution for modules like `jquery__event_type_pointer` :(

narqo:
I think one should use solutions like https://github.com/azproduction/autopolyfiller for such things

PR #502, opened 2014-04-17
dom: Wrong result of isFocusable() for link without href but with tabindex attribute

### LINKED ISSUE #501: dom: Wrong result of isFocusable() for link without href but with tabindex attribute

narqo:
fix #501 

\cc @dfilatov


dfilatov:
lgtm

Issue #54, opened 2013-06-11
Review i-bem.bemhtml


veged:
cc @veged

qfox:
ping @veged

veged:
since new bem-xjst it has become irrelevant

PR #508, opened 2014-04-26
Refactroing i-bem.bemhtml

narqo:
see #54 

cc @veged @indutny @dfilatov


indutny:
LGTM, if tests are passing. Though, I have no idea if the code style is really good :)

dfilatov:
Looks a little bit better than before )

PR #514, opened 2014-05-05
BEM.I18N: Use modules. Closes #59

### LINKED ISSUE #59: BEM.I18N: Use modules
Depend on #58.

cc @dfilatov

tadatuta:
Quite ugly but working.

Works the same as in `bem-bl` for `i18n.html` tech and via `modules.require(['i-bem__i18n'])` for `browser.js` and `node.js`. In templates `BEM.I18N` syntax is left for backward compatibility but it doesn't pollute global scope in browser or node.js bundle. Still I'm not sure it's a good idea to keep it like that.

To use `i18n` in BEMHTML within `node.js` bundle needed something like this

``` js
var vm = require('vm'),
    BEMHTML = require('fs').readFileSync('./desktop.bundles/index/index.bemhtml.js');

modules.define('BEMHTML', ['i-bem__i18n'], function(provide, I18N) {
    var ctx = vm.createContext({ BEM: { I18N: I18N } });
    vm.runInContext(BEMHTML, ctx);
    provide(ctx.BEMHTML);
});
```

but it's possible to write `i18n.bemhtml` tech to build bemhtml bundle with `i18n` code to use it with `require('index.bemhtml.js').BEMHTML`

Here's [project-stub branch](https://github.com/bem/project-stub/tree/i18n) with `i18n` to test.

@narqo @dfilatov @veged

### DIFF
--- a/.bem/techs/browser.js+bemhtml.js
+++ b/.bem/techs/browser.js+bemhtml.js
@@ -19,7 +19,7 @@ exports.techMixin = {
             });
     },
 
-    concatBemhtml: function(res, output, opts) {
+    getBemhtml: function(output, opts) {
         var _this = this,
             context = this.context,
             declaration = context.opts.declaration;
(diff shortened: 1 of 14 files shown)


narqo:
Current `BEM.I18N` from bem-bl is very slow, too Ya-specific, not reliable (if we're talking about Node.js and concurrency requests) and very complicated. I suggest you to plan a meeting after the holidays to talk about localisation from the the open-source point of view. This PR is too hacky :(

veged:
+1 to @narqo

PR #517, opened 2014-05-06
page: Save modes from page on page__body. Closes #516

### LINKED ISSUE #516: page: Save modes from page on page__body
Now it's impossible to use template for `page` and the only way is to write them for `page__body` which is far from being intuitive.

I suggest to save results of applying modes of the `page` within `page__body` context.

tadatuta:
cc @narqo @mishanga

PS: It works the same in BH already.

### DIFF
--- a/common.blocks/page/page.bemhtml
+++ b/common.blocks/page/page.bemhtml
@@ -20,7 +20,14 @@ block('page')(
                         ]
                     },
                     // Добавляем элемент, чтобы сработал другой шаблон и не было зацикливания
-                    this.extend(ctx, { elem : 'body' })
+                    this.extend(ctx, {
+                        elem : 'body',
+                        attrs : apply('attrs'),
+                        cls : apply('cls'),
+                        js : apply('js'),
+                        mix : apply('mix'),
+                        content : apply('content')
+                    })
                 ]
             }
         ]);


tadatuta:
any considerations?

narqo:
Could you provide a peace of real BEMJSON where all this staff would be useful?

tadatuta:
that's not about `bemjson`, it's about writing `bemhtml`-templates for `page` block:

``` js
block('page').js()(true);
```

won't work right now.

tadatuta:
@narqo so what do you think?

common.blocks/page/page.bemhtml
```diff
@@ -20,7 +20,14 @@ block('page')(
...
                     },
                     // Добавляем элемент, чтобы сработал другой шаблон и не было зацикливания
-                    this.extend(ctx, { elem : 'body' })
+                    this.extend(ctx, {
+                        elem : 'body',
+                        attrs : apply('attrs'),
+                        cls : apply('cls'),
+                        js : apply('js'),
+                        mix : apply('mix'),
+                        content : apply('content')
```
narqo, line 29: I have a bad feelings about this thing (not only this actually). It feels like `this.ctx.content`'s templates would be processed twice in some situations, since you have `apply('content')` here and `applyCtx()` above.

narqo, line 29: Maybe we could rewrite page's template so it will work as one expected?

narqo:
As far as I understand, with your example:

``` js
block('page').js()({ a : 1 });
```

I will have HTML like this:

``` html
<body class="page__body i-bem" data-bem="{ 'page__body' : { 'a' : 1 } }">
```

But it looks very strange for me

tavriaforever:
@narqo if you will write 

``` javascript
block('page').js()({ a : 1 });
```

You will have html like this:

``` html
<body class="page">
```

tadatuta:
@narqo nope, final html will be just as expected but i'm not sure if template will be processed twice

narqo:
> nope, final html will be just as expected [..]

Yep, it's a magic! ;)

Maybe we could simplify this template like this:

``` diff
diff --git a/common.blocks/page/page.bemhtml b/common.blocks/page/page.bemhtml
index f39f93a..93997ff 100644
--- a/common.blocks/page/page.bemhtml
+++ b/common.blocks/page/page.bemhtml
@@ -1,6 +1,8 @@
 block('page')(

-    def()(function() {
+    def().match(function() { return !this._defPageApplied; })(function() {
+        this._defPageApplied = true;
+
         var ctx = this.ctx;
         applyCtx([
             ctx.doctype || '<!DOCTYPE html>',
@@ -19,27 +21,22 @@ block('page')(
                             ctx.favicon? { elem : 'favicon', url : ctx.favicon } : ''
                         ]
                     },
-                    // Добавляем элемент, чтобы сработал другой шаблон и не было зацикливания
-                    this.extend(ctx, { elem : 'body' })
+                    ctx
                 ]
             }
         ]);
+
+        this._defPageApplied = false;
     }),

-    elem('body')(
-        tag()('body'),
-        content()(function() {
-            return [
-                applyNext(),
-                this.ctx.scripts
-            ];
-        }),
-        def()(function() {
-            // Обратно очищаем поле elem, чтобы сохранить правильный контекст
-            this.ctx.elem = null;
-            applyNext();
-        })
-    ),
+    tag()('body'),
+
+    content()(function() {
+        return [
+            applyNext(),
+            this.ctx.scripts
+        ];
+    }),
```

In this case additional templates for `page` works fine.

P.S. It seems, that `this._defPageApplied` guard won't be needed with `bem-xjst@0.6.x`

tadatuta:
Closed in honour of https://github.com/bem/bem-core/issues/542

PR #520, opened 2014-05-08
#519 Update template

ilyar:


.bem/techs/bemhtml.js
```diff
@@ -62,6 +62,7 @@ exports.techMixin = {
 
         vars.ElemName && tmpl.push('.elem(\'{{bemElemName}}\')');
         vars.ModVal && tmpl.push('.' + (vars.ElemName ? 'elemMod' : 'mod') + '(\'{{bemModName}}\', \'{{bemModVal}}\')');
+        tmpl.push('()');
```
tadatuta, line 65: I've got 2 arguments to left it as is:
1. If there's no template for the block it's better to remove it. So when compilation fails it's like a kind of linting saying "there's something wrong with your code, go and fix".
2. there are a lot of templates with just one custom mode and it's convenient to write them in "dot-notation": `block('bla').js()(true);` or `block('bla').content()(/* ... */);` so having parenthesis will disturb.

PR #521, opened 2014-05-08
i-bem__dom: Prevent block to be reinitialized on the DOM node, that has been processed with destructor (close #518)

### LINKED ISSUE #518: i-bem__dom: Prevent block to be reinitialized on the DOM node, that has been processed with destructor
Imaging a BEM-tree like this:

``` html
<outer-block>
  <inner-block/>
</outer-block>
```

When we're removing `outer-block` with `BEMDOM.destruct(outerBlock.domElem)`, there could be a situation when `outer-block`'s destructor has been processed already, but during `inner-block`'s destruct, it emits some events that could lead to reinitialization process on the `outer-block`'s `domElem`, that is still in the DOM in this moment. After that `outer-block`'s domElem would be removed from the DOM, but new block's instance 
would "dangle" in browser's memory :(

See live demo: http://jsbin.com/jobal/1/edit?js,console

There are two `zooo → __constructor` statements in console, but with different `_uniqId`. And there is only one `zooo → destruct`

dfilatov:


narqo:
@dfilatov could you check my spec for this case?

dfilatov:
@narqo spec looks good

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -905,7 +907,7 @@ var DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
      * @param {Boolean} [excludeSelf=false] Exclude the main domElem
      */
     destruct : function(ctx, excludeSelf) {
-        findDomElem(ctx, BEM_SELECTOR, excludeSelf).each(function(i, domNode) {
+        reverse.call(findDomElem(ctx, BEM_SELECTOR, excludeSelf)).each(function(i, domNode) {
```
narqo, line 910: Wow, invoking `Array.prototype.reverse` with the context of jQuery object  return jQuery object. I haven't knew that. Cool! :)

narqo:
:+1:

PR #530, opened 2014-05-19
i-bem: enable to use `false` in subscription to mod change event (close #529)

### LINKED ISSUE #529: i-bem: Unable to use `false` in subscription to mod change event
``` js
this.liveInitOnBlockInsideEvent({ modName: 'checked', modVal: false }, 'button', this._onChecked);
```

если модвал фолс

dfilatov:


dfilatov:
cc @narqo

narqo:
:+1:

PR #542, opened 2014-06-03
page: Simplify template to save modes on page__body. Closes #516

### LINKED ISSUE #516: page: Save modes from page on page__body

tadatuta:
just applied @narqo's patch from https://github.com/bem/bem-core/pull/517#issuecomment-43195972

### DIFF
--- a/touch.blocks/page/page.bemhtml
+++ b/touch.blocks/page/page.bemhtml
@@ -1,7 +1,7 @@
 block('page')(
-    mode('head')(function() {
+    elem('head').content()(function() {
         return [
-            { block : 'ua' },
+            applyNext(),
             {
                 elem : 'meta',
                 attrs : {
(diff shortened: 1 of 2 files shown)


tavriaforever:
cool

narqo:
If this one works as expected (dev-mode, production, redefinition?), than :fire:

tadatuta:
@narqo Added fix for touch template, which was broken after https://github.com/bem/bem-core/commit/789125c086dd73e58973593826a2518b9be97fb5

// cc @mishanga

touch.blocks/page/page.bemhtml
```diff
@@ -13,12 +13,13 @@ block('page')(
...
-            this.ctx.head
+            { elem : 'link', attrs : { name : 'apple-mobile-web-app-capable', content : 'yes' } }
         ];
     }),
 
     mix()(function() {
-        return applyNext().concat({ block : 'ua', js : true });
+        var mix = applyNext() || [];
+
+        return [].concat(mix, { block : 'ua', js : true });
```
narqo: Maybe we could optimize it a little bit?

``` js
var mix = applyNext(),
  uaMix = [{ block : 'ua',.. }];
return mix? uaMix.concat(mix) : uaMix;
```

tadatuta: @narqo done

narqo:
> If this one works as expected (dev-mode, production, redefinition?), than :fire:

+1

PR #567, opened 2014-06-30
Fix pointer events

### LINKED ISSUE #545: jquery__event_type_pointer: pointerup doesn't bubble on iOS
``` html
<html>
<body>
  <div>пойнтрапмне</div>
</body>
</html>
```

``` js
$('body').on('pointerup', function(e) {
  console.log(e.type, e.target, e.currentTarget);
});
```

Tapping in iOS device (at least iOS 7) doesn't print anything in console. 

At the same time, this works:

``` js
$('div').on('pointerup', function(e) {
  console.log(e.type, e.target, e.currentTarget);
});
```

and prints:

```
pointerup  <div>…</div>  <div>…</div>
```

narqo:
@dfilatov @veged do you want to review changes :)

closes #545


narqo:
See related PR in bem-components bem/bem-components#737 to tests

common.blocks/jquery/__event/_type/jquery__event_type_pointer.spec.js
```diff
@@ -4,9 +4,18 @@ modules.define(
     function(provide, $, sinon) {
 
 describe('jquery__event_type_pointer', function() {
+    var elem;
+
+    beforeEach(function() {
+        elem = $('<div/>').appendTo('body');
```
narqo, line 10: note here: as pointer events use native events which are listened on `document` we can't test them with the detached DOM node

common.blocks/jquery/__event/_type/jquery__event_type_pointerpressrelease.deps.js
```diff
@@ -1,7 +1,6 @@
...
-    mustDeps : {
-        block : 'jquery', elems : {
-            elem : 'event', mods : { type : 'pointer' }
-        }
-    }
-})
+({
+    mustDeps : {
+        block : 'jquery',
+        elem : 'event', mods : { type : 'pointer' }
```
veged: looks like `pointer` deps `pointerpressrelease` and `pointerpressrelease` deps `pointer` — think should keep just `pointerpressrelease` deps `pointer`

also you removed deps on `jquery` itself (difference between `elems` and `elem`)

common.blocks/jquery/__event/_type/jquery__event_type_pointer.js
```diff
@@ -1,553 +1,840 @@
...
 /**
- * Normalize touch-event by keeping all the
- * possible properties normalized by jQuery.
- *
- * @see http://api.jquery.com/category/events/event-object/
+ * Basic pointer events polyfill
  *
- * @param {Object} e event
+ * @see https://dvcs.w3.org/hg/pointerevents/raw-file/tip/pointerEvents.html
+ * @author Kir Belevich
```
veged: is it still keep some from Kir's version?

narqo: The [repo](https://github.com/narqo/jquery-pointerevents) is my fork of his project, so I decided to leave `@author`. But actually I don't care

common.blocks/jquery/__event/_type/jquery__event_type_pointerpressrelease.deps.js
narqo: > think should keep just pointerpressrelease deps pointer

But our own codebase from `bem-component` project would hurt. E.g. https://github.com/bem/bem-components/blob/v2/common.blocks/button/button.deps.js#L8

IMHO, this should go to major changes

veged: we can move `pointer` to some `pointerpolyfill` (without `pointerpressrelease` and `pointerclick` deps) and keep `pointer` just as "API" for including all pointer events — this will keep changes in minor manner

PR #588, opened 2014-07-09
page.bemhtml: Remove ieN support. Closes #379

### LINKED ISSUE #379: page.bemhtml: Remove ieN.css support
There's no more need to support ie6 and ie7, so we don't need shortcut for `ieN.css` as it saves maximum one line in bemjson but require all ie6-9.css to be built.

tadatuta:


dfilatov:
:shipit:

Issue #594, opened 2014-07-12
BEMTREE error handling


apsavin:
I think, we should stop use

``` javascript
Vow
        .allResolved(ctx.__queue)
        .always(function() {
            return ctx._buf;
        });
```

[here](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree#L98)
and start use something like

``` javascript
Vow
        .all(ctx.__queue)
        .then(function() {
            return ctx._buf;
        });
```

instead.
It seems to me, that it's impossible to do error handling now in BEMTREE, because it always behaves like all is ok.
What do you think about it?

tadatuta:
+1
// cc @narqo

narqo:
I don't mind

veged:
### Meeting results:
- fix `apply` (https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree#L100) to return result AND promises (so we need to use `spread` after, need issue for document it)
- add 3 more optional parameters to `doAsync` (`onFulfilled`, `onRejected`, `context`) — like in `when` (need issue for document it, maybe same as in prev point)

PR #595, opened 2014-07-13
Impove bemtree error handling

apsavin:
...so you can do local error handling inside blocks and global inside process,  which calls BEMTREE.apply.
[Issue 594](https://github.com/bem/bem-core/issues/594)

### DIFF
--- a/common.blocks/i-bem/i-bem.bemtree
+++ b/common.blocks/i-bem/i-bem.bemtree
@@ -96,8 +96,8 @@ exports.apply = BEMContext.applyAsync = function BEMContext_applyAsync(context) 
     var ctx = new BEMContext(context || this, oldApply);
     ctx._buf = ctx.apply();
     return Vow
-        .allResolved(ctx.__queue)
-        .always(function() {
+        .all(ctx.__queue)
+        .then(function() {
             return ctx._buf;
         });
 };
--- a/common.blocks/i-bem/i-bem.bemtree
+++ b/common.blocks/i-bem/i-bem.bemtree
@@ -67,27 +67,51 @@ BEMContext.prototype.generateId = function generateId() {
     return this.identify(this.ctx);
 };
 
-BEMContext.prototype.doAsync = function doAsync(fn) {
+/**
+ *
+ * @param {Function} fn
+ * @param {...*} [args]
+ * @returns {vow:Promise}
+ */
+BEMContext.prototype.doAsync = function doAsync(fn, args) {
+    var callArgs = arguments.length > 1 && Array.apply(null, arguments),
+        promise;
+
+    promise = callArgs ? Vow.invoke.apply(Vow, callArgs) : Vow.invoke(fn);
+
+    return this.addToQueue(this.wrapPromise(promise));
+};
+
+/**
+ * @param {vow:Promise} promise
+ * @returns {vow:Promise}
+ */
+BEMContext.prototype.addToQueue = function addToQueue(promise) {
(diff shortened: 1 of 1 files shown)


apsavin:
It's not so easy as it seems. May be, we should do something like this (it's a sketch).
And then use like this:

``` javascript
block('async-block').content()(function () {
    return this.addToQueue(
       this.wrapPromise(doRequestThatWillReturnAPromise())
             .then(function (data) {
                     //all is  ok
              },
              function (data) {
                     //we can do error handling here, if we want to handle the error in bemtree
                     //or we can do not provide this callback and then error handling must be done outside bemtree
               },
               this)
       );
});
```

By the way, is there any tests for bemtree?

apsavin:
Guys, I think it's a serious problem and I don't know how to use BEMTREE in production without proper error handling.
Can you give me some feedback?

apsavin:
@tadatuta?.. @narqo?.. Anybody?..

tadatuta:
@apsavin I understand the reason of the first commit but I'm not sure if we really need to make things more complicated as in second one.

Isn't it enough to handle errors in BEMTREE like this:

``` js
    return this.doAsync(function() {
        var defer = Vow.defer(),
            promise = defer.promise();

        setTimeout(function() {
            defer.reject('упячка хеппенс');
        }, 10);

        return promise
            .fail(function(err) {
                return err; // will set the content of b1 to 'упячка хеппенс'
            });
    });
});
```

apsavin:
I'm afraid, no.
See the example:

``` javascript
    return this.doAsync(function() {
        return doAsyncStuff().then(function (data) {
             return applyCtx({
                  block: 'b2', 
                  content: data,
                  b1Stuff: this.ctx.b1Stuff //will fail. We need to do applyAsync to have this.ctx here 
             }); 
        }, function (err) {return err;}, this);
    }, this);
});
```

``` javascript
    return this.doAsync(function() {
        return doAsyncStuff();
    }).then(function (data) {
             return applyCtx({
                  block: 'b2', 
                  content: data,
                  b1Stuff: this.ctx.b1Stuff 
             }); 
        }, function (err) {
            return err; //We will never be here because of changed error handling in doAsync
        }, this);;
});
```

Am I wrong?
I don't want to say that `wrapPromise` and `addToQueue` are good names for all that stuff, it's just a point to start a discussion.

apsavin:
@tadatuta I created more simple variant [here](https://github.com/bem/bem-core/pull/601)

apsavin:
Guys, can you give me a date, when you'll be able to pay attention to this problem?

tadatuta:
@apsavin, it looks like BEMTREE needs to be totally reviewed as besides issue with error handling there's also another one with nested async calls: we need to do something like [this](https://github.com/bem/project-stub/commit/ada414d805d054571c7539539eebe6aa36146173#diff-f189455bb77673c7adc84822cf7a8919R11) in order to make them work now.

@veged, maybe we can schedule skype meeting together with @apsavin to find out what to do?

veged:
we can do it at this week — pls arrange one of our "O2" meetigns for

(we also need @narqo to come)

tadatuta:
@apsavin, as I can see in @veged's schedule it's possible to meet at 30.07 or 31.07 from 12:00 till 15:00 and 01.08 from 13:00 till 15:00. if it suits you, please, add me on skype (I'm `tadatuta` there) and tell which time is best.

apsavin:
We decided to use another way #594

Issue #596, opened 2014-07-13
More arguments for doAsync bemtree method


apsavin:
Right now only one argument can be passed into [doAsync](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree#L70) inside [bemtree](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree) tech files and this argument is function to invoke. I think, we should implement possibility to pass any number of arguments. It will not very hard, because doAsync [uses](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree#L77) Vow.invoke internally, which [can](https://github.com/dfilatov/vow/blob/master/lib/vow.js#L892) behaves like this.

I need it, for example, for things like this:

``` javascript
this.doAsync(function(_this){
    return request.get(route)
        .then(function () {/*ok*/}, function() {/*error*/}, _this);
}, this);
```

Of course, I can do right now like this:

``` javascript
this.doAsync(function(){
    return request.get(route);
}).then(function () {/*ok*/}, function() {/*error*/}, this);
```

But it's not the same, especially if [pull-request](https://github.com/bem/bem-core/pull/595) about error handling will be merged.

tadatuta:
sounds reasonable, can you please suggest PR as well?

apsavin:
We decided to use doAsync arguments in the different manner #594

PR #597, opened 2014-07-13
Improve doAsync bemtree function: any number of arguments

apsavin:
[Issue 596](https://github.com/bem/bem-core/issues/596)


apsavin:
We decided to use doAsync arguments in the different manner #594

Issue #583, opened 2014-07-06
i-bem, i-bem-dom: `findElem` should update elem's cache


qfox:
Can `findElem` method rewrite `elem`'s internal cache if called with the same parameters?

aristov:
// cc @dfilatov

veged:
Yes, it can do this in case of simple params. Feel free to make PR.

qfox:
Related issue #598

qfox:
@veged izi.

PR #598, opened 2014-07-15
Make elem method a little bit smarter

L0stSoul:
Now we have some problem with very aggressive elem caching e.g
bemjson:

```
{
    block: 'test',
    js: true,
    content:  {
        elem: 'el',
        content: 'first'
    }
}
```

.js:

```
{
modules.require(['i-bem__dom'], function(DOM) {
    DOM.decl('test',{
        onSetMod: {
            'js': function() {
               console.log(this.elem('el').text()); // OK - first
               DOM.update(this.domElem, '<div class="test__el">second</div>');
               console.log(this.elem('el').text()); // WTF - first
            }
        }
        });
});
}
```

So it's reach field for mistake, or we must use findElem in all of the cases for simplify things.

I think this change have a little impact on performance, and may make things better. May be :)

### DIFF
--- a/common.blocks/i-bem/__dom/i-bem__dom.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.js
@@ -758,7 +758,7 @@ var DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
         var key = name + buildModPostfix(modName, modVal),
             res;
 
-        if(!(res = this._elemCache[key])) {
+        if(!(res = this._elemCache[key]) || !res.parent().length) {
             res = this._elemCache[key] = this.findElem(name, modName, modVal);
             res.__bemElemName = name;
         }


narqo:
Guess that `this.dropElemCache('el')` is more suitable for such cases?

L0stSoul:
I think such things need to be done automatically. may be:

```
this.dropElemCache(elems)
```

can be done somwhere inside update method? because  it's too complicated for such common operation. I often saw a bugs because elem cache was obsolete and not updated. And again often saw a developers who use findElem in such kind of situation just because it's error-free.

dfilatov:
In bem-core@v3 we try to get rid of `dropElemCache`

L0stSoul:
good to hear.

aristov:
@L0stSoul we have already done some tuning about cache in #583. Guess other changes within v2 branch may break backwards compatibility. Could you check does this problem still exists in v3?

PR #601, opened 2014-07-23
Improve error handling and content mode

apsavin:
Variant of BEMTREE error hangling
[Issue 594](https://github.com/bem/bem-core/issues/594)
so you can use:

```
    return this.doAsync(function() {
        return doAsyncStuff();
    }).then(function (data) {
             return applyCtx({
                  block: 'b2', 
                  content: data,
                  b1Stuff: this.ctx.b1Stuff 
             }); 
        }, function (err) {
            return err; 
        }, this);;
});
```

or even:

```
    return this.doAsync(oAsyncStuff()).then(function (data) {
             return applyCtx({
                  block: 'b2', 
                  content: data,
                  b1Stuff: this.ctx.b1Stuff 
             }); 
        }, function (err) {
            return err; 
        }, this);;
});
```

Some part of old doAsync logic is in content mode now. I think, it's more convenient.

### DIFF
--- a/common.blocks/i-bem/i-bem.bemtree
+++ b/common.blocks/i-bem/i-bem.bemtree
@@ -3,7 +3,6 @@
 oninit(function(exports) {
 
 var undef,
-    BEM_ = {},
     toString = Object.prototype.toString,
     isArray = Array.isArray || function(obj) {
         return toString.call(obj) === '[object Array]';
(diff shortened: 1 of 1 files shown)


tadatuta:
@apsavin, looks good to me. thank you!

@veged, @narqo ?

narqo:
I don't understand, how and why this would work. @tadatuta have you checked it?

@apsavin do the case below would still work?

```
block('b1').def()(function() {
  return this.doAsync(function() { return 1 });
});
```

I don't understand how my promise would appear in `this.__queue`.

apsavin:
No, of course, it will not work, it's for content mode only. Do we really need `doAsync` in default mode?

narqo:
It would be a very strange API that works only in "some specific cases". In my opinion we really should avoid it. `doAsync` should work properly, regardless the context in which it used.

And yes we used to use `doAsync` not only for content. 

Imaging that you have some `global` block that provides global parameters for the whole BEM-tree. It could be something similar to http://www.polymer-project.org/docs/polymer/polymer.html#global

apsavin:
Well, then we have two ways:
1. implement logic from content mode (put promise into queue) in every mode (sounds not very cool)
2. use separate method for that operation (like [here](https://github.com/bem/bem-core/pull/595))

@narqo can you suggest another way? What do you think about it?

apsavin:
We decided to use another way #594

PR #609, opened 2014-07-27
i-bem__dom: base `live` is not called in derived block (close #608)

### LINKED ISSUE #608: i-bem__dom: base `live` is not called in derived block
- `block1` is declared with `live` section
- `block1` processes its live section (DOM.init)
- `block2` is declared as derived from `block1` with `live` section contained base call within
- `block2` processes its live section

Unexpected result: base call within `live` section of `block2` doesn't work

dfilatov:
cc @veged @narqo

### DIFF
--- a/common.blocks/i-bem/__dom/i-bem__dom.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.js
@@ -925,9 +925,17 @@ var DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
         if('live' in this) {
             var noLive = typeof res === 'undefined';
 
-            if(noLive ^ heedLive) {
+            if(noLive ^ heedLive) { // should be opposite to each other
                 res = this.live() !== false;
-                this.live = functions.noop;
+
+                var blockName = this.getName(),
+                    origLive = this.live;
+
+                this.live = function() {
+                    return this.getName() === blockName?
+                        res :
+                        origLive.apply(this, arguments);
+                };
             }
         }
 
--- a/common.blocks/i-bem/__dom/i-bem__dom.spec.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.spec.js
@@ -811,6 +811,88 @@ describe('i-bem__dom', function() {
         });
     });
 
+    describe('live', function() {
+        it('should properly use live of base block', function() {
+            var spy1 = sinon.spy(),
+                spy2 = sinon.spy();
+
+            DOM.decl('block1', {}, { live : spy1 });
+            DOM.decl({ block : 'block2', baseBlock : 'block1' }, {}, {
+                live : function() {
+                    this.__base.apply(this, arguments);
+                    spy2();
+                }
+            });
+
+            DOM.init(BEMHTML.apply([
+                { block : 'block1', js : true },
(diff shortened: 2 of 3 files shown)


common.blocks/i-bem/__dom/i-bem__dom.spec.js
```diff
@@ -811,6 +811,89 @@ describe('i-bem__dom', function() {
...
+        it('should properly use live after adding declaration', function() {
+            var spy1 = sinon.spy(),
+                spy2 = sinon.spy();
+
+            DOM.decl('block1', {}, { live : spy1 });
+
+            DOM.init(BEMHTML.apply({ block : 'block1', js : true }));
+
+            spy1.should.have.been.calledOnce;
+            spy2.should.not.have.been.called;
```
narqo: This assert looks useless, isn't it?

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -927,7 +927,15 @@ var DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
 
             if(noLive ^ heedLive) {
```
narqo: May be we could add some notes about `^` here? ;)

common.blocks/i-bem/__dom/i-bem__dom.spec.js
dfilatov: Yes, you're definitely right. Removed.

common.blocks/i-bem/__dom/i-bem__dom.js
dfilatov: I haven't changed this code )
Could you suggest reasonable comment?

narqo: Come on, it's your code, I don't understand it :P

dfilatov: it's equivalent to: `if((noLive && !heedLive) || (!noLive && heedLive)) {`

veged: `if(noLive ^ heedLive) { // should be always different from each other` ;-)

dfilatov: added

narqo:
Looks good

common.blocks/i-bem/__dom/i-bem__dom.spec.js
```diff
@@ -811,6 +811,88 @@ describe('i-bem__dom', function() {
...
+                    spy2();
+                }
+            });
+
+            spy1.should.have.been.calledOnce;
+            spy2.should.have.been.calledOnce;
+
+            DOM.init(BEMHTML.apply({ block : 'block1', js : true }));
+
+            spy1.should.have.been.calledOnce;
```
veged, line 889: what the difference with `spy1.should.have.been.calledTwice` in pref test?

dfilatov, line 889: In previous spec two different blocks (`block1`, `block2` which derived from `block1`) are used.
In this spec `block1` is declared twice.

veged:
:+1:

PR #617, opened 2014-07-29
`pointerpress`/`pointerrelease` event should be fired if only left mouse button is pressed

### LINKED ISSUE #607: `pointerpress`/`pointerrelease` event should be fired if only left mouse button is pressed

narqo:
fixes #607 

backport #616 onto v3


dfilatov:
:+1:

PR #627, opened 2014-08-02
Update bem-xjst to 0.8.1

tadatuta:
Port from `v2`
https://github.com/bem/bem-core/issues/625


tadatuta:
Closed in favour of #642

Issue #602, opened 2014-07-23
Provide possibility to extend BEMTREE


apsavin:
I want to add some stuff inside BEMTREE, for example, additional method to BEMContext prototype. As I can see, it's impossible right now. What about exporting xjstApply and BEMContext?
Imagine:

```
BEMTREE.BEMContext.prototype.doRequest = function () {};
BEMTREE.apply(ctx);
```

I can send a pull request.)

tadatuta:
@apsavin 
if `doRequest` is some external module it's possible to put it as global variable with the help of `require('vm').runInContext()` or if it's some helper function which you are to call in a lot of blocks you can put it into `this` in runtime:

``` js
mode('')(applyNext({
    doRequest: function() {}
}))
```

does it solve you goal?

apsavin:
`require('vm').runInContext()` is not available in the browser.
And what about performance? I think it's better to put doRequest in prototype at once, not to extend `this` every time.

apsavin:
@tadatuta, I need something like this [here](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemtree#L94): 

``` javascript
// Export xjst's apply and BEMContext for extensibility
exports._xjstApply = exports.apply;
exports._BEMContext = BEMContext;
// Wrap xjst's apply and export our own
exports.apply = BEMContext.applyAsync = function BEMContext_applyAsync(context) {
    var ctx = new exports._BEMContext(context || this, exports._xjstApply);
    ctx._buf = ctx.apply();
    return Vow
        .allResolved(ctx.__queue)
        .always(function() {
            return ctx._buf;
        });
};
```

I think, it's important to provide such extensibility in data templates, so frameworks and apps will be able to add their own sugar and helpers.
It is a common pattern in BEMTREE: 

``` javascript
block('b1').content(function () {
    return this.doAsync(function () {
        return doRequest();
    }).then(function (data) {return data;});
});
```

And it's ugly. I want to use just

``` javascript
block('b1').content(function () {
    return this.doRequest().then(function (data) {return data;});
});
```

And it's not the BEMTREE's business to know anything about how I want to do request, it's business of my framework or my application. The access to BEMContext and xjst's apply will provide all the possibilities.
And what if I want to have something in `BEMContext` instances, that depends on `http.IncomingMessage`, for example? It seems to me that only possibility to redefine `apply` and to extend/redefine `BEMContext` will help me.)

apsavin:
[...]

PR #636, opened 2014-08-12
Possibility to extend BEMTREE

apsavin:
For #602, as discussed. Just like in BEMHTML.


tadatuta:
LGTM, thanks!

@Murdalay, I think we should add some notes about possibility to extend context to `BEMHTML` and after release of this update also to `BEMTREE` docs.

``` js
oninit(function(exports, context) {
    context.BEMContext.prototype.newMethod = function() {
        /* ... */
    }
});

    this.newMethod();
});
```

tadatuta:
@veged guess, we can merge?

PR #644, opened 2014-08-16
Improve BEMTREE error handling

apsavin:
As discussed #594

### DIFF
--- a/common.blocks/i-bem/i-bem.bemtree
+++ b/common.blocks/i-bem/i-bem.bemtree
@@ -3,10 +3,12 @@
 oninit(function(exports) {
 
 var undef,
-    BEM_ = {},
     toString = Object.prototype.toString,
     isArray = Array.isArray || function(obj) {
         return toString.call(obj) === '[object Array]';
+    },
+    isFunction = function(obj) {
+        return typeof obj === 'function';
     },
     buildEscape = (function() {
         var ts = { '"' : '&quot;', '&' : '&amp;', '<' : '&lt;', '>' : '&gt;' },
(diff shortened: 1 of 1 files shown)


apsavin:
@tadatuta @veged any feedback?

common.blocks/i-bem/i-bem.bemtree
```diff
@@ -67,27 +69,57 @@ BEMContext.prototype.generateId = function generateId() {
...
-        promise = Vow.invoke(fn);
+        restore = function(response) {
+            this._mode = mode;
+            this.ctx   = ctx;
+            this.block = block;
+            this.elem  = elem;
+            this.mods  = mods;
+            this.elemMods = elemMods;
+            return response;
+        }.bind(this),
```
veged: I suggest to avoid `.bind()` — just

``` js
var _this = this;
function restore(response) {
    _this._mode = mode;
    ...
}
```

I know that it was before ;-)

common.blocks/i-bem/i-bem.bemtree
```diff
@@ -67,27 +69,57 @@ BEMContext.prototype.generateId = function generateId() {
...
+        promise = isFunction(fn) ? Vow.invoke(fn, context) : fn;
+
+    if(onFulfilled) {
+        if(onRejected) {
```
veged: can be done with ternary if: `promise = onRejected? ... : ...;`

veged:
I add some inline comments. Also, what about docs or at least issue for them?

common.blocks/i-bem/i-bem.bemtree
apsavin: I think, it's more clean and readable without ternary operator, but as you wish.

common.blocks/i-bem/i-bem.bemtree
apsavin: ok

tadatuta:
@veged Let's merge?

veged:
wait a minute... the problem is that this changes are not back compatible (am I right? we need to add `.spread()` to `BEMTREE.apply()` usage) — so we need to follow SemVer

and also it can not be shipped without docs #658 

am I right? @narqo @dfilatov we need v3?

aristov:
@apsavin could you port this PR onto v3? And fix @veged 's comments please.

apsavin:
I'll try to find time. All @veged's comments were fixed.

PR #671, opened 2014-09-07
Update documentation

4exova:
Modifiers' change events and description of block inheritance were added.


tadatuta:
вошло в https://github.com/bem/bem-core/pull/696

PR #689, opened 2014-10-07
Documentation and jsDoc fixes, update enb config

sipayRT:


sipayRT:
@tadatuta

.enb/make.js
```diff
@@ -34,17 +34,14 @@ module.exports = function(config) {
     config.includeConfig('enb-bem-examples');
     config.includeConfig('enb-bem-docs');
     config.includeConfig('enb-bem-specs');
-    config.includeConfig('enb-bem-tmpl-specs');
```
tadatuta, line 37: why've you removed tmpl-specs? there are at least `page` templates in `bem-core`

common.blocks/i-bem/i-bem.ru.md
```diff
@@ -0,0 +1,937 @@
+Блок `i-bem` — это блок-хелпер, позволяющий создавать другие блоки.
```
tadatuta, line 1: this text is outdated

common.blocks/page/page.ru.md
```diff
@@ -0,0 +1,151 @@
+﻿Блок `page` создаёт теги верхнего уровня страницы:
```
tadatuta, line 1: this text needs review

common.blocks/page/page.ru.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Тег <meta> с указанием кодировки
+
+`BEMHTML`:
+
+```js
+content: [
+{
+    tag: 'meta',
+    attrs: { 'http-equiv': 'content-type', content: 'text/html; charset=utf-8' }
```
tadatuta, line 71: `attrs: { charset: 'utf-8' }`

.enb/make.js
sipayRT, line 37: because we haven't any tmpl-specs in this repository. i think that we should to add new technologies just when we really use them. if you need tests for `page` block, I can write them

common.blocks/i-bem/i-bem.ru.md
sipayRT, line 1: I just renamed `wiki` to `md` and changed the syntax. should we create issue for updating this documentation?

common.blocks/page/page.ru.md
sipayRT, line 1: i think this text is outdated too

common.blocks/i-bem/i-bem.ru.md
tadatuta, line 1: yes, let's create issues for all the things which are not right

.enb/make.js
tadatuta, line 37: yes, it's better to write tests and not to remove them from config

sipayRT:
@tadatuta tmpl-specs for `page` was added. But when I started to write tmpl-specs for `ua`, I found some error in BH-templates. it's look like bug in BH. I'll make an example to show this error

sipayRT:
https://github.com/bem/bh/issues/98

mishanga:
Fixed in bem/bh#100, published in https://github.com/bem/bh/releases/tag/v3.1.4

sipayRT:
tnx!

PR #695, opened 2014-10-13
Adding English version of BEMJSON guide

rasulksirov:
BEMJSON guide in English - differences from the Russian version include (as per email discussion with @tadatuta):
- In **BEMJSON and templating in bem-core**, the NB note - elaborating on what kind of arbitrary data is normally fed to the BEMTREE template engine, to align with the previous statement about BEMJSON being the input data format for that template engine;
- In **HTML representation** subsection - removing the NB note about how BEMHTML template priority is considered in template authoring on the basis that such details would be more relevant in the BEMHTML guide.


tadatuta:
@rasulksirov Rasul, as this pull request is just about adding one new file, I think it'd be better to squash all commits to one.

The translation itself looks good, thanks!

rasulksirov:
Closing this pull request, will open a new one with a single commit of the relevant file.

PR #697, opened 2014-10-14
Correct urls

gela-d:
@tadatuta @tormozz48 
On our bem-info this links are incorrect

### DIFF
--- a/common.docs/reference/reference.ru.md
+++ b/common.docs/reference/reference.ru.md
@@ -18,7 +18,7 @@
 * JavaScript;
 * CSS;
 * БЭМ;
-* [BEMJSON](http://ru.bem.info/libs/bem-core/2.2.0/templating/bemjson).
+* [BEMJSON](http://ru.bem.info/technology/bemjson/current/bemjson/).
 
 
 **В документе не описаны** настройка среды разработки и процедуры компиляции шаблонов, синтаксис BEM-XJST-шаблонизаторов, синтаксис BEMJSON.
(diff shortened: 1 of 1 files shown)


common.docs/reference/reference.en.md
```diff
@@ -1179,7 +1179,7 @@ Template's definition in the mode `js` (sub-predicate `js()`) makes sense if onl
 
 **See also**:
 
-  * [JS-implementation of the block i-bem](http://bem.info/libs/bem-bl/current/desktop/i-bem/)
+  * [JS-implementation of the block i-bem](http://bem.info/libs/bem-bl/dev/desktop/i-bem/)
```
tadatuta: why?

common.docs/reference/reference.ru.md
```diff
@@ -352,7 +352,7 @@ JavaScript.
 
 **См. также**:
 
-  * [JS-реализация блока i-bem](http://ru.bem.info/libs/bem-bl/current/desktop/i-bem/)
+  * [JS-реализация блока i-bem](http://ru.bem.info/libs/bem-bl/dev/desktop/i-bem/)
```
tadatuta: ?

common.docs/reference/reference.en.md
gela-d: I don't know :)
Reverted

common.docs/reference/reference.ru.md
gela-d: Because we can :)
Because we want the same urls https://github.com/bem/bem-core/pull/697/files#diff-7688330737dca987ee217c02ea0bb7c0R1182

tadatuta: you've just reverted that url ;)

PR #700, opened 2014-10-17
i-bem__dom: findBlocksInside may return blocks which aren't inited (close #699)

### LINKED ISSUE #699: i-bem__dom: findBlocksInside may return blocks which aren't inited

dfilatov:
сс @narqo @veged


narqo:
I'm not sure, but from [this peace of code](https://github.com/bem/bem-core/blob/cddfca7bc2417e2bb24770c34f1db7d0ebd44ba4/common.blocks/i-bem/__dom/i-bem__dom.js#L122-L130) I have a feeling that there are cases when `initBlock()` returns `undefined`?

dfilatov:
@narqo in all of these cases `forceLive=true` and `initBlock()` has no any chances not to return instance.

narqo:
OK

PR #739, opened 2014-11-26
Updating English versions of some documents

rasulksirov:
The English versions of the BEMTREE template engine guide and BEMJSON guide have been updated following Ami's comments and suggestions regarding the translation. The English version of the BEMTHML template engine guide has been reviewed, proofread and its content adjusted as per the Russian version.


tadatuta:
thanks!

PR #740, opened 2014-11-27
Links to external docs updated, "Russian version only" note inserted where applicable

rasulksirov:
Sorry, forgot to update the URLs in the documents in the previous commit. Have now reviewed all the links to external documents and updated where appropriate: replaced with the English version if one now exists for the linked documentation, and added a "Russian version only"-type comment wherenever an explicit reference to another document is made (such as the "See also" note).

### DIFF
--- a/common.docs/bemtree-reference/bemtree-reference.en.md
+++ b/common.docs/bemtree-reference/bemtree-reference.en.md
@@ -10,36 +10,36 @@ The guide describes:
 * Input data processing and BEMJSON generation;
 * BEMTREE-based solutions to some typical problems.
 
-**The target audience for this guide** are web developers and HTML coders who use the [BEM methodology](http://bem.info/method/).
+**The target audience for this guide** are web developers and HTML coders who use the [BEM methodology](https://bem.info/method/).
 
 The reader is assumed to be familiar with:
 
 * HTML
 * JavaScript
 * CSS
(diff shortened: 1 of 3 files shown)


common.docs/bemtree-reference/bemtree-reference.en.md
```diff
@@ -335,7 +336,7 @@ All context fields can be divided into two categories:
 
 **See also**:
 
-  * [Context](http://ru.bem.info/technology/bemhtml/current/templating/#context)
+  * [Context](http://ru.bem.info/technology/bemhtml/current/templating/#context)(Russian version only)
```
tadatuta: space before `(Russian version only)` needed

tadatuta:
Just a few common notes about links in documentation:
1. For all version-related links: we have virtual `current` version which resolves to last available version so if there's no particular need to specify version it's better to use it like this:
   Before: `[BEMJSON](http://en.bem.info/technology/bemjson/2.3.0/bemjson/)`
   After: `[BEMJSON](http://en.bem.info/technology/bemjson/current/bemjson/)`
2. It's better to use `https` for any link to `bem.info` (we're going to switch on redirect from http to https soon).

rasulksirov:
Thanks for the comments, I've now applied the suggested conventions to links in all three translated guide documents, link formatting should now be consistent across the three of them.

tadatuta:
thank you!

PR #743, opened 2014-11-29
i-bem.bemhtml: apply mode=js for mixed items (close #280)

### LINKED ISSUE #280: i-bem.bemhtml: apply mode=js for mixed items
Right now only `js` field from input context is considered.
But we need also to apply by mode=js for mixed item cause js params can be defined in bemhtml templates.

aristov:
resolves #280


common.blocks/i-bem/i-bem.bemhtml
```diff
@@ -384,11 +384,12 @@ def()(function() {
...
                                 mixItem.elemMods || mixItem.mods);
 
-                            if(mixItem.js) {
-                                (jsParams ||
-                                        (jsParams = {}))[BEM_INTERNAL.buildClass(mixBlock, mixItem.elem)] = mixItem.js === true?
-                                    {} :
-                                    mixItem.js;
+                            var mixItemJs = apply('js', { block : mixBlock, elem : mixElem });
+                            mixItemJs = mixItemJs?
+                                this.extend(mixItem.js, mixItemJs === true? {} : mixItemJs) :
```
narqo, line 392: What if will explicitly convert arguments passed to `extend` to objects with `Object(o1)`? It seems that we have a lot such checks in our code

aristov, line 392: #747

veged:
you also forgot about https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.bemhtml#L340

and please try to add some tests:
- in bemjson `js: true` + in bemhtml `js: { ... }`
- in bemjson `js: { ... }` + in bemhtml `js: true`

narqo:
Don't forget to think about implementing it in bh 
//cc @mishanga

aristov:
For me: squash and merge, add issue to `bh`.

mishanga:
В BH на уровне движка такого не будет. Можно явно в нужных местах делать `ctx.process(ctx.mix()).js`.

PR #746, opened 2014-11-30
ua: add support for ie11+

persidskiy:
Hey!
IE since 11 version doesn't match previous regexp. Moreover, it simulates useragent like FF 11.

```
IE 11: "Mozilla/5.0 (Windows NT 6.3; Trident/7.0; .NET4.0E; .NET4.0C; rv:11.0) like Gecko"
FF 35: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:35.0) Gecko/20100101 Firefox/35.0" 
```

Thus, our current implementation of `ua` recognizes IE 11 as FF 11.

//cc @dfilatov @narqo

### DIFF
--- a/desktop.blocks/ua/ua.js
+++ b/desktop.blocks/ua/ua.js
@@ -11,6 +11,7 @@ var ua = navigator.userAgent.toLowerCase(),
         /(webkit)[ \/]([\w.]+)/.exec(ua) ||
         /(opera)(?:.*version|)[ \/]([\w.]+)/.exec(ua) ||
         /(msie) ([\w.]+)/.exec(ua) ||
+        /(msie)(?:.*? rv:([\w.]+))/.exec(ua.replace('trident', 'msie')) ||
         ua.indexOf('compatible') < 0 && /(mozilla)(?:.*? rv:([\w.]+)|)/.exec(ua) ||
         [],
     matched = {


desktop.blocks/ua/ua.js
```diff
@@ -11,6 +11,7 @@ var ua = navigator.userAgent.toLowerCase(),
         /(webkit)[ \/]([\w.]+)/.exec(ua) ||
         /(opera)(?:.*version|)[ \/]([\w.]+)/.exec(ua) ||
         /(msie) ([\w.]+)/.exec(ua) ||
+        /(msie)(?:.*? rv:([\w.]+))/.exec(ua.replace('trident', 'msie')) ||
         ua.indexOf('compatible') < 0 && /(mozilla)(?:.*? rv:([\w.]+)|)/.exec(ua) ||
```
persidskiy, line 15: Also, why do we need the pipe symbol after `rv:([\w.]+)` ?

narqo, line 15: This code has been taken from [jquery-migrate](https://github.com/jquery/jquery-migrate/blob/cadf8fcbabe85307c8e5a67b542fbc2028200142/src/core.js#L57) project

persidskiy:
So, what about merge?

desktop.blocks/ua/ua.js
narqo, line 15: To be honest, I don't like this `ua.replace()` statement here. Can't we use `(msie|trident)(...)` and than check

``` js
if(browser.trigent) browser.msie = true
```

narqo, line 15: Another strange thing I have noticed is "IE11 for the desktop on 64-bit Windows 8.1 with enterprise mode enabled" (WAT?) from [User-agent string changes](http://msdn.microsoft.com/en-us/library/ie/hh869301%28v=vs.85%29.aspx) of MSDN:

```
Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 6.1; Trident/4.0; Tablet PC 2.0)
```

It has "trident" token, but there is no "rv" part :( 

The same shit is with several other IE 11 x64 from the article above.

persidskiy, line 15: > Can't we use `(msie|trident)(...)` and than check...

No, because old regexp with `(msie)` token rely on number version after this token.

> IE11 for the desktop on 64-bit Windows 8.1 with enterprise mode enabled ...

It will be recognized as IE8, because it has `MSIE 8.0` substring. 

So, this behavior was before my changes and I won't break something.

dfilatov:
@Coltspb Is it needed for real case? Or just for consistency?

persidskiy:
I had a task at SERP, where it was needed and I decided to propose it here.

There was a problem with our new popup, which twitches on scroll if his anchor is in fixed block like header. It happens only in IE and some safari and it's ugly if popup is big enough, like big suggest in SERP. Thus, I needed to sniff IE and safari, but `$.browser` and this `ua` block recognize IE 11 as FF 11.
Thus, this PR is here :)

desktop.blocks/ua/ua.js
narqo, line 15: Ok, but why not to add 

``` js
/(trident)(?:.*? rv:([\w.]+))/.exec(ua)
```

and have a `if` check as with `chrome + webkit`.

narqo, line 15: @Coltspb ping

narqo:
@persidskiy ping!

aristov:
Reopen if needed.

PR #749, opened 2014-12-01
lint: jscs+jscs-jsdoc update

qfox:
About #586


qfox:
@veged @dfilatov Guys, what you think if we move all jscs rules to separate preset? Can you take a look at code here?

qfox:
@veged @dfilatov Need your review, guys.

veged:
@zxqfox what the purpose for moving all jscs rules to separate preset? other code seems like fixing spaces and looks good for me

qfox:
The idea in making linting library to have the same code style in all block libraries. I'm not sure which preset is the closest to the current code base. But we can make our own jshint+jscs presets as plugin or npm package. Like that: https://github.com/bem/bem-core/commit/29c6c47 (usign [zxqfox/jscs-bem](https://github.com/zxqfox/jscs-bem)) or like that [bem-site/bem-site-codestyle](https://github.com/bem-site/bem-site-codestyle). I've dropped my commit with that for now and will restore it if we need it.

The main thing here is that all functions marked as `@protected` in jsdoc will be reported on `npm run lint`.

When we will have a separate package with code style configuration for bem libraries? ;-)

veged:
@zxqfox can't catch point for separate packages — is it not enough just copy-paste small config file? ;-)

anyway — I have nothing against such presets functionality if it supported by tools

qfox:
@veged That's a long story ;-) Thanks!

Issue #782, opened 2014-12-29
i-bem__dom_elem-instances: Seems like a bug with 2 elements inside block


qfox:
Shortly: If we make 2 elements inside a block and call them then properties of first elem will share with the second one. It wasn't shared with block object so I think it's bug in `elemInstances`.

https://gist.github.com/zxqfox/45c35ac395dfec326555#file-elem-inst-test-browser-js-L46 and https://gist.github.com/zxqfox/45c35ac395dfec326555#file-output-L9

So. Is it a bug?

qfox:
It's unexpected and unclear behavior and definitely not a bug. Do we need some tests like that?

narqo:
To be honest, I don't know anything about elem-instances of `v2`.

/cc @aristov

qfox:
Closing this since elemInstances is experimental thing in v2

PR #783, opened 2014-12-29
i-bem_elem-instances: tests for decl

qfox:
Ref #782


aristov:
I think, that these tests are redundant. Look explanation here: https://gist.github.com/zxqfox/45c35ac395dfec326555

Issue #792, opened 2015-01-14
Pointer(press/release) don't work in IE8


sipayRT:
You can check it on https://github.com/bem/bem-core/tree/v2/common.blocks/jquery/__event/_type/jquery__event_type_pointerpressrelease.tests (don't forget to add es5-shim for IE8). In normal browsers you will see messages in console after clicking on the button, but not in IE8.

sipayRT:
@tadatuta

sipayRT:
@narqo

hcodes:
Fix #797, #798

PR #797, opened 2015-01-16
Pointer(press/release) don't work in IE8 #792

hcodes:


narqo:
:+1:

narqo:
BTW It seems, that `e.which` is not a standard property (see [article on MDN](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent.which)). Have tested it in all supported browsers?

narqo:
Oh, it's an jquery event object. Seems that it should be normalized

PR #802, opened 2015-01-18
jquery__pointernative: pointerenter / pointerleave events bubble up to the document root

### LINKED ISSUE #801: jquery__pointernative: pointerenter / pointerleave events bubble up to the document root
As `pointerenter` and `pointerleave` are equivalents of native `mouseenter` and `mouseleave` events respectively, they both must not bubble up through the DOM. 

See [`mouseleave` event description](https://developer.mozilla.org/en-US/docs/Web/Events/mouseleave) for reference.

narqo:
fix #801

/cc @dfilatov

### DIFF
--- a/common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
+++ b/common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
@@ -24,11 +24,12 @@ if(typeof modules === 'object' && modules.isDefined('jquery')) {
 var doc = document,
     USE_NATIVE_MAP = window.Map && window.Map.prototype.forEach,
     HAS_BITMAP_TYPE = window.MSPointerEvent && typeof window.MSPointerEvent.MSPOINTER_TYPE_MOUSE === 'number',
-    POINTERS_FN = function() { return this.size };
+    POINTERS_FN = function() { return this.size },
+    jqEvent = $.event;
 
 // NOTE: Remove jQuery special fixes for pointerevents – we fix them ourself
-delete $.event.special.pointerenter;
-delete $.event.special.pointerleave;
+delete jqEvent.special.pointerenter;
+delete jqEvent.special.pointerleave;
 
 /*!
  * Returns a snapshot of inEvent, with writable properties.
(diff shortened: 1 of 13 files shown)


narqo:
Note, the only browsers I've tested it are Chrome and Fx.

dfilatov:
Could you add spec for `pointerleave`?

narqo:
@dfilatov we already had simple tests for `pointerleave / pointerleave` events. But we didn't test event bubbling. 

:up:

narqo:
BTW: It seems that no one of currently popular Pointer Events Polyfils (MS, Polymer or jquery's PEP) would pass [this test case](https://github.com/bem/bem-core/pull/802/files#diff-810fc4229b7cb8db80e663abf226a289R160), as they use native `bubbling = false` and dispatch the event on the last nested target only  :metal:

common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
```diff
@@ -258,15 +268,23 @@ var dispatcher = {
...
-        if(!this.contains(event.target, event.relatedTarget)) {
-            this.enter(event);
+        this.enterLeave(event, this.enter);
+    },
+
+    enterLeave : function(event, fn) {
+        var target = event.target,
+            relatedTarget = event.relatedTarget;
+
+        while(target && !this.contains(target, relatedTarget)) {
```
dfilatov: Isn't it expensive? `contains` has inner cycle as well

narqo: Hm, yep, haven't thought about it :( 

What if I rewrite it like this:

``` js
if(!this.contains(target, relatedTarget)) {
  do {
    event.target = target;
    fn.call(this, event);  // we need to emit non bubbling events here on the subtree from target to relatedTarget
  } while(target = target.parentNode && target !== relatedTarget)
}
```

?

dfilatov: imho looks better

narqo:
:up:

common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
```diff
@@ -258,15 +268,25 @@ var dispatcher = {
...
     },
 
     enterOver : function(event) {
         this.over(event);
-        if(!this.contains(event.target, event.relatedTarget)) {
-            this.enter(event);
+        this.enterLeave(event, this.enter);
+    },
+
+    enterLeave : function(event, fn) {
```
narqo, line 279: I don't like the name of the method, but I couldn't decide how to call it :(

dfilatov:
:+1:

dfilatov:
@narqo Don't forget do the same PR to v3

PR #805, opened 2015-01-18
i-bem.bemhtml: Do not repeat block class on mix with same block's mod

### LINKED ISSUE #804: i-bem.bemhtml: Do not repeat block class on mix with same block's mod
``` js
{
    block: 'b1',
    mix: { block: 'b1', mods: { m1: 'v1' } }
}
```

results with

``` html
<div class="b1 b1 b1_m1_v1"></div>
```

It's inconsistent with `BH` and creates issues with tmpl-specs.

tadatuta:
closes #804


tadatuta:
The first commit was kinda reasonable but it broke [this case](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.test.bemhtml/files/i-bem/gh-232.bemhtml).
So I added second one with fix and a test case for current issue but now solution looks ugly :(

Anyways there is a couple issues with mixes which we need to discuss first:
- https://github.com/bem/bem-core/issues/806
- https://github.com/bem/bem-core/issues/807

// cc @narqo @veged

ruslansagitov:
:up:

qfox:
I would say need more tests. :-\

ruslansagitov:
I need it in [bem-bl](https://github.com/bem/bem-bl).

veged:
@tadatuta go ahead! merge it and port to v3

PR #809, opened 2015-01-19
loader_type_bundle: append loaded css to the end of head (close #808)

### LINKED ISSUE #808: loader_type_bundle: append loaded css to the end of head
Otherwise `mustDeps` in loaded css is ignored.

dfilatov:
cc @narqo


narqo:
Currently there is no "true" solution for CSS bundle injection. Both variants have there drawbacks as new styles from bundle can pollute (or even affect) common styles from the exact page. 

One possible solution was described in [this comment](https://github.com/veged/borschik/issues/31#issuecomment-8947906) from borschik, but it requires a lot of additional tooling :(

My :+1: here, as currently we don't have any real widgets, like `mooa` from the internals which will suffer in any cases ;)

PR #838, opened 2015-01-29
package: bump vow version

### LINKED ISSUE #837: vow: Bump version

qfox:
Fixes #837

### DIFF
--- a/common.blocks/vow/vow.vanilla.js
+++ b/common.blocks/vow/vow.vanilla.js
@@ -1,7 +1,7 @@
 /**
  * @module vow
  * @author Filatov Dmitry <dfilatov@yandex-team.ru>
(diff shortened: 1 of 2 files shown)


dfilatov:
@zxqfox Primarily, we're going to bump version here — https://github.com/bem/bem-core/blob/v2/common.blocks/vow/vow.vanilla.js. Could you do it as well?

qfox:
@dfilatov Lol. sorry. Do we need commit with package.json?

dfilatov:
@zxqfox I think we can include it as well

dfilatov:
@zxqfox Thanks! Could you do the same PR to v3?

qfox:
@dfilatov Yep, sure. #839

PR #842, opened 2015-01-30
i-bem__dom: update _elemCache on findElem

### LINKED ISSUE #583: i-bem, i-bem-dom: `findElem` should update elem's cache

qfox:
Fixes #583

### DIFF
--- a/common.blocks/i-bem/__dom/i-bem__dom.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.js
@@ -736,12 +736,22 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
             modName = undef;
         }
 
+        names = names.split(' ');
+
         var _self = this.__self,
-            selector = '.' +
-                names.split(' ').map(function(name) {
-                    return _self.buildClass(name, modName, modVal);
-                }).join(',.'),
+            keys = names.map(function(name) {
+                return _self.buildClass(name, modName, modVal);
+            }),
+            isSingleName = keys.length === 1,
+            selector = '.' + (isSingleName? keys[0] : keys.join(',.')),
             res = findDomElem(ctx, selector);
+
+        // caching results if possible
+        (ctx.length === 1 && this.domElem.length === 1 && ctx[0] === this.domElem[0]) &&
+            keys.forEach(function(key, i) {
+                (this._elemCache[key] = isSingleName? res : res.filter('.' + key))
+                    .__bemElemName = names[i];
+            }, this);
 
         return strictMode? this._filterFindElemResults(res) : res;
     },
@@ -768,15 +778,8 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
      * @returns {jQuery} DOM elements
      */
     _elem : function(name, modName, modVal) {
-        var key = name + buildModPostfix(modName, modVal),
-            res;
-
-        if(!(res = this._elemCache[key])) {
-            res = this._elemCache[key] = this.findElem(name, modName, modVal);
-            res.__bemElemName = name;
-        }
-
-        return res;
+        return this._elemCache[this.__self.buildClass(name, modName, modVal)] ||
+            this.findElem(name, modName, modVal);
     },
 
     /**
--- a/common.blocks/i-bem/__dom/i-bem__dom.spec.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.spec.js
@@ -291,7 +291,7 @@ describe('i-bem__dom', function() {
 
         afterEach(function() {
             DOM.destruct(rootNode);
-            delete DOM.blocks['b-root'];
+            delete DOM.blocks['root'];
             delete DOM.blocks['b1'];
         });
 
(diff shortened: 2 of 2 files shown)


common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -723,13 +723,21 @@ var DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
...
             selector = '.' +
-                names.split(' ').map(function(name) {
+                names.map(function(name) {
                     return _self.buildClass(name, modName, modVal);
                 }).join(',.'),
             res = findDomElem(ctx, selector);
 
+        (ctx === this.domElem) && names.forEach(function(name) {
+                var key = _self.buildClass(name, modName, modVal),
+                    elem = this._elemCache[key] = res.filter('.' + key);
```
veged: In most cases we find a single elem (`names.length === 1`). So we can avoid unnecessary `res.filter`:

``` js
var isSingleName = names.length === 1;
...
elem = this._elemCache[key] = isSingleName? res : res.filter('.' + key);
```

qfox: Sounds reasonable, thanks!

qfox: Is there difference between `name + buildModPostfix(modName, modVal)` and `_self.buildClass(name, modName, modVal)` ?

veged: definitely — `name + buildModPostfix(modName, modVal)` more like violation of encapsulation of building full class

qfox:
What's with coverage?

narqo:
(just in case...): you can investigate the coverage decreasing [directly on Coveralls](https://coveralls.io/builds/1833468)

qfox:
I didn't see anything about `common.blocks/i-bem/__dom/i-bem__dom.js` but see `common.blocks/i-bem/i-bem.vanilla.js` change.

narqo:
Oh, sh.., looks like we have a fail positive here. Take a look at the build logs in Travis: there are few failed specs.

/cc @tadatuta @andrewblond We got a critical problem with enb-bem-specs here!

qfox:
Fixed bug.
Made a little refactoring.
Squashed commits.

Need review ;-)

https://github.com/enb-bem/enb-bem-specs/issues/21

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -736,13 +736,23 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
...
         }
 
+        names = names.split(' ');
+
         var _self = this.__self,
-            selector = '.' +
-                names.split(' ').map(function(name) {
-                    return _self.buildClass(name, modName, modVal);
-                }).join(',.'),
+            keys = names.map(function (name) {
```
narqo: code style problem here

qfox: Sorry, fixed. How it doesn't caught on pre-commit?..

blond:
> Oh, sh.., looks like we have a fail positive here. Take a look at the build logs in Travis: there are few failed specs.

@zxqfox, @narqo https://github.com/bem/bem-core/issues/908

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -736,13 +736,23 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
...
-                }).join(',.'),
+            keys = names.map(function(name) {
+                return _self.buildClass(name, modName, modVal);
+            }),
+            isSingleName = keys.length === 1,
+            selector = '.' + (isSingleName? keys[0] : keys.join(',.')),
             res = findDomElem(ctx, selector);
 
+        // caching results if possible
+        (ctx === this.domElem) && keys.forEach(function(key, i) {
```
narqo: As far as I remember, it's not save to compare jQuery-nodes in this way. Check this:

``` js
var link = $('.link:eq(0)');
link === $(link);   // ☚ false
```

Can you compare underlying DOM-nodes instead?

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -736,13 +736,23 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
...
+                var elem = this._elemCache[key] = isSingleName? res : res.filter('.' + key);
+                delete elem.prevObject; // drop link to original set after filtering
```
dfilatov: What is `prevObject`? I can't find any usage of it.

narqo: Can you point me, where this `prevObject` key came from?

common.blocks/i-bem/__dom/i-bem__dom.js
```diff
@@ -768,15 +778,8 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
...
-        var key = name + buildModPostfix(modName, modVal),
-            res;
-
-        if(!(res = this._elemCache[key])) {
-            res = this._elemCache[key] = this.findElem(name, modName, modVal);
-            res.__bemElemName = name;
-        }
-
-        return res;
+        var key = this.__self.buildClass(name, modName, modVal);
```
narqo: Looks like this `key` variable can be inlined now?

zxqfox: `res.filter` for sure.

common.blocks/i-bem/__dom/i-bem__dom.js
qfox: From `res.filter` for sure.

common.blocks/i-bem/__dom/i-bem__dom.js
qfox: Yes, if you want.

common.blocks/i-bem/__dom/i-bem__dom.js
qfox: You're right. Need to fix that.

common.blocks/i-bem/__dom/i-bem__dom.js
qfox: Inlined.

common.blocks/i-bem/__dom/i-bem__dom.js
qfox: Fixed, added test. Check please.

There is `.is` function but I'm not sure about we should use it. It's pretty slow in compare with raw equal (but still much more readable): http://jsperf.com/jquery-objects-comparison/9

dfilatov:
@zxqfox I'm still puzzled how `prevObject` is used.

qfox:
@dfilatov Idea was to remove `res` (placed in `prevObject` key) to reduce links count and make possible to collect garbage. I can drop it if you don't like it.

dfilatov:
I open i-bem__dom.js with your changes and try to find substring "prevObject".
There's only occurrence:
`delete elem.prevObject;`

qfox:
It comes from here: https://github.com/jquery/jquery/blob/2380028ec4a6a77401b867a51de26a3cb8e8d311/src/core.js#L70

dfilatov:
Oh, my god. Let's not to use jquery internals.

qfox:
Okay, but `res` will never be collected ;-)

dfilatov:
Why?

qfox:
Because `res.filter` create new object for each `name`/`key` and `res` is stored in `prevObject` property in each new object. Nvm.

dfilatov:
If I lose link to `res`, they will be collected, won't they?

qfox:
That's why I tried to delete it. in other way it will be stored in `prevObject` property. Where I'm wrong? I can't get ;-)
As an opposite there is another idea to make empty chain (`$()`) and push there filtered elements. But is it worth?

PR #857, opened 2015-02-01
Fix for event.Emitter.emit method Close #855

### LINKED ISSUE #855: events.Emitter.emit() method ignores data field
If the second argument is provided to `events.Emitter.emit()` it just do nothing instead of supplying it value to e.data field.

Murdalay:


Murdalay:
Maybe there's an other way to pass the e.data value with `event.Emitter.emit`?

dfilatov:
See details in #855

PR #864, opened 2015-02-03
jquery__event_type_winresize: ie -> msie

### LINKED ISSUE #862: jquery__event_type_winresize: Fix typo
@dfilatov Dima, we have https://github.com/bem/bem-core/blob/v2/desktop.blocks/jquery/__event/_type/jquery__event_type_winresize.js in `bem-core` but it looks like it's not used anyway. Can't we just remove it?

tadatuta:
Closes #862


narqo:
:fire: — squash and merge!

PR #869, opened 2015-02-06
page__conditional-comment: Implement template

tadatuta:
1. Updated `enb-bem-tmpl-specs` to check HTML comments
2. Fixed old reference HTMLs after [1]
3. Implemented templates as discussed in #551 but with two additions:
   - `includeOthers` param to generate CC which works in specified IE versions and all other browsers
   - possibility to specify versions both with `lte` and `<=` syntax
4. Wrote basic examples

cc @sipayRT @hcodes @veged

### DIFF
--- a/desktop.blocks/page/__conditional-comment/page__conditional-comment.bemhtml
+++ b/desktop.blocks/page/__conditional-comment/page__conditional-comment.bemhtml
@@ -1,0 +1,23 @@
+block('page').elem('conditional-comment')(
+    tag()(false),
+
+    content()(function() {
+        var ctx = this.ctx,
+            cond = ctx.condition
+                .replace('<', 'lt')
+                .replace('>', 'gt')
+                .replace('=', 'e'),
+            hasNegation = cond.indexOf('!') > -1,
+            includeOthers = ctx.msieOnly === false,
+            hasNegationOrIncludeOthers = hasNegation || includeOthers;
+
+        return [
+            '<!--[if ' + cond + ']>',
+            includeOthers? '<!' : '',
+            hasNegationOrIncludeOthers? '-->' : '',
+            applyNext(),
+            hasNegationOrIncludeOthers? '<!--' : '',
+            '<![endif]-->'
+        ];
+    })
+);
--- a/desktop.blocks/page/__conditional-comment/page__conditional-comment.ru.md
+++ b/desktop.blocks/page/__conditional-comment/page__conditional-comment.ru.md
@@ -1,0 +1,31 @@
+```javascript
+({
+    block : 'page',
+    title : 'page__conditional-comments',
+    head : [
+        {
+            elem : 'conditional-comment',
+            condition : '<= IE 8',
+            content : { elem : 'css', url : '_page.ie.css' }
+        },
+        {
+            elem : 'conditional-comment',
+            condition : '! IE',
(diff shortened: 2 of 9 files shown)


sipayRT:
@tadatuta  Do we need to remove [these](https://github.com/bem/bem-core/tree/v2/desktop.blocks/page/__css) templates for block `page`?

tadatuta:
@sipayRT they should be already removed in v3 and we can't remove them in minor version anyway

sipayRT:
ok

sipayRT:
@hcodes @veged ping

qfox:
@sipayRT https://github.com/bem/bem-core/blob/v3/desktop.blocks/page/__css/page__css.bh.js  404. ;-) just for sure.

sipayRT:
@zxqfox v3? this PR to v2

qfox:
@sipayRT @tadatuta said "they should be already removed". For sure: they ARE already removed. ;-)

sipayRT:
@zxqfox ok, tnx

desktop.blocks/page/__conditional-comment/page__conditional-comment.bemhtml
```diff
@@ -0,0 +1,23 @@
...
+    tag()(false),
+
+    content()(function() {
+        var ctx = this.ctx,
+            cond = ctx.condition
+                .replace('<', 'lt')
+                .replace('>', 'gt')
+                .replace('=', 'e'),
+            hasNegation = cond.indexOf('!') > -1,
+            includeOthers = ctx.includeOthers,
```
narqo: Let's rename `includeOther=true` to `msieOnly=false`

desktop.blocks/page/__conditional-comment/page__conditional-comment.ru.md
```diff
@@ -0,0 +1,31 @@
...
+            elem : 'conditional-comment',
+            condition : '> IE 8',
+            includeOthers : true,
+            content : 'For IE9+ and all other browsers'
+        }
+    ],
+    scripts : [
+        {
+            elem : 'conditional-comment',
+            condition : 'lte IE 8',
```
narqo, line 26: Why `lte` and not `<=`?

narqo:
lgtm, but fix the notes please

desktop.blocks/page/__conditional-comment/page__conditional-comment.ru.md
tadatuta, line 26: this's a draft of documentation. an example with `<=` is above.

desktop.blocks/page/__conditional-comment/page__conditional-comment.bemhtml
tadatuta: @narqo 
should be switched on by default?
I mean is `msieOnly` equal `true` if not set `false` explicitly? or we should treat `undefined` as `false` as well?

narqo: I think that default value should be `false`, so the behaviour shouldn't change. I mean 

``` js
includeOthers = ctx.msieOnly === true
```

tadatuta:
@narqo up

narqo:
Rebase, merge and port to v3

PR #876, opened 2015-02-11
Usage of jscs-bem and code style fixes for @v2

### LINKED ISSUE #875: Move jscs rules to preset @v2
Refs: https://github.com/bem/bem-components/issues/1220, https://github.com/bem/bem-components/pull/1221

qfox:
Ref bem/bem-components#1221
Fixes #875


narqo:
@zxqfox can you provide more meaningful title for the pull request

qfox:
@narqo There is an issue with a lot of info and meaningful title, but fixed.

common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
```diff
@@ -532,7 +532,7 @@ var touchEvents = {
     removePrimaryPointer : function(pointer) {
         if(pointer.isPrimary) {
             this.firstTouch = null;
-            //this.firstXY = null;
+            // this.firstXY = null; // no reason to left comments with code like this without a note
```
narqo: Can you make a comment here, like this:

``` js
// TODO(@narqo): It seems that, flushing `firstXY` flag explicitly in `touchmove` handler is enough.
// Original code from polymer doing this on every `removePrimaryPointer` call, but looks 
// like it is harmful in some of our usecases.
```

.jscs.json
```diff
@@ -15,28 +18,8 @@
         "common.blocks/ecma/__array/ecma__array.spec.js",
         "common.blocks/jquery/__event/_type/jquery__event_type_pointerclick.js",
         "common.blocks/jquery/__event/_type/jquery__event_type_pointer.js",
-        "common.blocks/jquery/__event/_type/jquery__event_type_pointerpressrelease.js"
+        "common.blocks/jquery/__event/_type/jquery__event_type_pointerpressrelease.js",
```
narqo: Looks like this `jquery__event_type*` section is obsoleted. 

At present, only `jquery__event_type_pointerclick` is a vendor plugin and should not be checked. `jquery__event_type_pointer` was renamed to `jquery__event_type_pointernative`, so we can remove this line; as well ass `jquery__event_type_pointerpressrelease`. It has been rewritten as a part of bem-core, so it no longer a vendor plugin.

common.blocks/jquery/__event/_type/jquery__event_type_pointernative.js
qfox: Instead of existing one, or as addition?

narqo: Instead of `// no reason to left comments with code like this without a note`

qfox: :+1:

narqo:
And please, port it to v3 after it'd been done.

qfox:
It's already in v3: https://github.com/bem/bem-core/commit/a70a8f154d53fc926ba5cc584b12a4f5085e444e

narqo:
Sweet! Than ping me please when you'll finish here

qfox:
@narqo ping ;-)

Issue #892, opened 2015-03-03
i-bem__dom: add shortcut for `this.elem(elemName, boolModName, true)` as `this.elem(elemName, boolModName)`


dfilatov:

qfox:
That's strange because better to find all elems with name `elemName` and which has mod `boolModName` with any value by `this.elem(elemName, boolModName)`.

dfilatov:
@zxqfox I disagree with you. We've already have the same behaviour for `setMod(elem, boolModName)`. Moreover, I can't even contemplate any reasonable case to find all the elems with any mod value, but I have enough cases to find elem with boolean mod.

qfox:
@dfilatov Wait, but if you looking for boolean mod then you won't check it's value, right?

dfilatov:
@zxqfox I'll check that value === true.

qfox:
@dfilatov I just want to understand what's the difference. I don't see a big difference between checking for true, and just existing. But still if you want to say that `elem(name, mod, [val=true])` is more intuitive than `elem(name, mod, [val=undefined])` — I'm not sure that you're right. Even if you have a lot of cases to find your elems with boolean mods. Right?

qfox:
Btw, checking for mod existence will totally cover searching of boolean mods but still will be useful for non-boolean mods. Reverse is not true.

dfilatov:
@zxqfox 
Is it clear for you that `setMod(modName)` is a shortcut for `setMod(modName, true)`, not for `setMod(modName, undefined)`?
There's a difference between checking for true and just existing when you're finding elem by css selector.
`.block__elem_modName` is a simplest selector, which is heavily optimized in the browsers. There's no so simple selector for your case.

qfox:
@dfilatov Actually, nope. But if you insist — I'll try to apply it for myself as clear. It will be hard but I'll try.

About searching by css selectors: Am I right that we loading all dom elements with `i-bem` class so we have links to them all in memory? Am I right that we know which ones has modifiers? If so should we depend on css selectors? How we can find non-dom blocks with modifiers? That's smells like inconsistent hell for me. ;-(

qfox:
Stupid idea. We can add both classes for modifiers if you insist on using css selectors: `.some_mod` and `.some_mod_value`, but use the one with value for js.

dfilatov:
[...]

PR #893, opened 2015-03-03
i-bem: setMod/hasMod convert non-boolean values to string

### LINKED ISSUE #890: i-bem: setMod should convert non-boolean value to string
``` js
block.setMod('blah', 2);
console.log(typeof block.getMod('blah'));  // number

block.setMod('blah', '2');
console.log(typeof block.getMod('blah'));  // string
```

Current behaviour looks weird and it seems that any value in the example above may fire several onSetMod-handlers (see [buildModFnName()](https://github.com/bem/bem-core/blob/6ce978f653cd84a7a490e654d6c6dbf378cb7546/common.blocks/i-bem/i-bem.vanilla.js#L53-L59)).

And the same goes to `hasMod`

narqo:
closes #890

/cc @dfilatov

### DIFF
--- a/common.blocks/i-bem/i-bem.vanilla.js
+++ b/common.blocks/i-bem/i-bem.vanilla.js
@@ -260,7 +260,7 @@ var BEM = inherit(events.Emitter, /** @lends BEM.prototype */ {
      * Checks whether a block or nested element has a modifier
      * @param {Object} [elem] Nested element
      * @param {String} modName Modifier name
-     * @param {String} [modVal] Modifier value
+     * @param {String|Boolean} [modVal] Modifier value. If defined and not of type String or Boolean, it is casted to String
      * @returns {Boolean}
      */
     hasMod : function(elem, modName, modVal) {
@@ -282,6 +282,11 @@ var BEM = inherit(events.Emitter, /** @lends BEM.prototype */ {
                 invert = true;
             }
         }
+
+        var typeModVal = typeof modVal;
+        typeModVal === 'string' ||
+            typeModVal === 'boolean' ||
+            typeModVal === 'undefined' || (modVal = modVal.toString());
 
         var res = this.getMod(elem, modName) === modVal;
         return invert? !res : res;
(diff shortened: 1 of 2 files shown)


dfilatov:
commit title doesn't correspond its content

dfilatov:
I have two points about that:
- jsdoc said that it took `String` param. So passing `Number` to it can't be treated as a bug.
- we have only special case —`Boolean`, so we should explicitly cast to `String` any other values, not just `Number`.

dfilatov:
@narqo ping

narqo:
> jsdoc said that it took String param. So passing Number to it can't be treated as a bug.

Agree with it. But at the same time I'd prefer to have a warnings (or an error) about such things. Currently it not that obvious, how [that code works](http://jsbin.com/jobal/3/watch?html,js,console).

> [..] so we should explicitly cast to String any other values, not just Number.

I think I haven't understood your point. Which other types in addition to String and Number do you mean? 

BTW, we've fixed jsdoc of these methods as well.

dfilatov:
> I think I haven't understood your point. Which other types in addition to String and Number do you mean?

I mean we don't need to explicitly consider `Number` as a special case. `Boolean` (and maybe `undefined`) is only special case. For any other values we should just call `.toString()`.

qfox:
Propose `{Boolean|*}` for type (or `{?}` if we want to do here #893). And description: `true or undefined would be treated as "true" string`.
But what about false?

dfilatov:
No, `undefined` should not be treated as `true`, especially as string, `setMod(modName, undefined)` is equivalent of deleting this mod. 
In case of `setMod(modName)` modVal should be treated as Boolean `true`.

qfox:
@dfilatov Sorry, I meant #892

dfilatov:
@zxqfox Do you ever sleep? )

qfox:
@dfilatov Should I? ;-)

narqo:
:up:

narqo:
@dfilatov pong

common.blocks/i-bem/i-bem.vanilla.js
```diff
@@ -283,7 +283,12 @@ var BEM = inherit(events.Emitter, /** @lends BEM.prototype */ {
             }
         }
 
-        var res = this.getMod(elem, modName) === modVal;
+        var typeModVal = typeof modVal;
+        typeModVal === 'string' ||
+            typeModVal === 'boolean' ||
+            typeModVal === 'undefined' || (modVal = modVal.toString());
+
+        var res = this.getMod(elem, modName) ===  modVal;
```
dfilatov: double space before `modVal`

dfilatov:
lgtm

dfilatov:
don't forget about v3, will you?

PR #904, opened 2015-03-07
i-bem: setMod/hasMod cast non-boolean mod values to string

narqo:
port #893 onto v3
part of #890

/cc @dfilatov

### DIFF
--- a/common.blocks/i-bem/i-bem.spec.js
+++ b/common.blocks/i-bem/i-bem.spec.js
@@ -159,6 +159,12 @@ describe('i-bem', function() {
                     .getMod('mod1')
                         .should.be.true;
             });
+
+            it('should cast non-boolean mod value to string', function() {
+                block
+                    .setMod('mod1', 0)
+                    .getMod('mod1').should.be.equal('0');
+            });
         });
 
         describe('delMod', function() {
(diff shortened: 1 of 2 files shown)


common.blocks/i-bem/i-bem.spec.js
```diff
@@ -187,14 +193,14 @@ describe('i-bem', function() {
                 block.hasMod('mod1').should.be.true;
             });
 
-            it('in short form should return true for empty mod\'s value', function() {
+            it('in short form should return false for empty mod\'s value', function() {
```
narqo, line 196: I'm not sure about why there was "true" if `shoud.be.false` is checked inside the test?

common.blocks/i-bem/i-bem.vanilla.js
```diff
@@ -257,12 +257,15 @@ var BemEntity = inherit(events.Emitter, /** @lends BemEntity.prototype */ {
     /**
      * Checks whether a BEM entity has a modifier
      * @param {String} modName Modifier name
-     * @param {String} [modVal] Modifier value
+     * @param {String|Boolean} [modVal] Modifier value. If not of type String or Boolean, it is casted to String
      * @returns {Boolean}
      */
     hasMod : function(modName, modVal) {
+        var withModVal = arguments.length === 2;
```
narqo: Can we decide that `hasMod('mod1', undefined)` is a useless edge case? At least until someone will catch this.

aristov: No, it's important.

narqo:
@dfilatov ping!

dfilatov:
lgtm

common.blocks/i-bem/i-bem.spec.js
qfox, line 196: Looks like copy-paste issues.

PR #906, opened 2015-03-11
The blocks description rework according to the new template

Murdalay:


Murdalay:
dom description yet still outdated. Don't comment it, please.

Murdalay:
Closed as it doubles the separate commits for docs.

PR #911, opened 2015-03-11
objects: extend fails in case of target=null (close #910)

### LINKED ISSUE #910: objects: extend fails in case of target=null

dfilatov:
cc @veged @narqo

### DIFF
--- a/common.blocks/objects/objects.spec.js
+++ b/common.blocks/objects/objects.spec.js
@@ -30,6 +30,11 @@ describe('objects', function() {
                 .should.be.eql({ p1 : 'v1' });
         });
 
+        it('should return new object if target is null', function() {
+            objects.extend(null, { p1 : 'v1' })
+                .should.be.eql({ p1 : 'v1' });
+        });
+
         it('should properly extend object with "hasOwnProperty" property', function() {
             objects.extend(
                 { hasOwnProperty : '' },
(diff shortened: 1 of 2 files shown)


narqo:
Looks good. Let's fix method's JSDoc as well

narqo:
Looks good. Let's fix method's JSDoc as well

dfilatov:
fix what? `typeof null === 'object'`

qfox:
:+1:

narqo:
@dfilatov I think it's good to have at least something like this is JSDoc:

```
@param {Object} target object to extend or null
```

dfilatov:
According to https://github.com/bem/bem-core/blob/issues/910%40v2/common.blocks/objects/objects.vanilla.js#L18 target can be anything, so I don't understand why `null` is the case which need to be described in jsdoc.

qfox:
-1 for noticing `null` in jsdoc. It's an overkill since `typeof null === object`.

As compromise I suppose to use:

```
@param {?Object} target object
```

https://developers.google.com/closure/compiler/docs/js-for-compiler — Nullable type

narqo:
@zxqfox at the same time they have explicit `null` in [@param example on the same page](https://developers.google.com/closure/compiler/docs/js-for-compiler#tag-param)

Anyway, I don't insist. Let's merge than

qfox:
@narqo okay ;-( But they have it in `{Type}`, not in description.

PR #912, opened 2015-03-11
page__conditional-comment: Implement template

tadatuta:
port of https://github.com/bem/bem-core/pull/869


narqo:
I believe there shouldn't be any pitfalls here. LGTM

PR #945, opened 2015-03-18
Update jquery__event.ru.md

apsavin:
Folks, it seems that there is a typo.


tadatuta:
@apsavin thank you! actually there's no need in `jquery` dependency at all

tadatuta:
// cc @Murdalay

PR #977, opened 2015-04-13
Rethink events

### LINKED ISSUE #394: i-bem-dom: Rethink events
Right now there are too many ways to use bem-events (directly and indirectly). We should rethink API.
### Current state
- `MyBlock.on('click', ...)` -- bind to all instances of `MyBlock`
- `MyBlock.on(domElem, 'click', ...)` -- bind to all instances of `MyBlock` inside or on `domElem`
- `AnotherBlock.liveInitOnBlockInsideEvent(MyBlock, 'click', ...)` -- bind to all instances of `MyBlock` inside all instances of `AnotherBlock`
- `AnotherBlock.liveInitOnBlockEvent(MyBlock, 'click', ...)` -- bind to all instances of `MyBlock` which placed on the same DOM-node as instances of `AnotherBlock`

veged:
close #394

TODO:
- [x] remove all old stuff (get rid of `EventEmitter` inheritance)
- [x] provide method for external bind/unbind (`bemDom.events`, `bemDom.domEvents`)
- [x] refactoring copy-paste between dom/bem event managers
- [x] add `.once` helper
  - [x] dom
  - [x] bem
- [x] specs for multi bind with same arguments
- [x] modifiers change events
  - [x] add support for modifiers change events
  - [x] add specs (don't forget to test destructing event propagation)
- [x] ~~refactor specs for `wrapSpy` everywhere~~
- [x] specs for live init on events
- [x] stopPropagation for bem events
- [x] make `emit()`, `events()` and `domEvents()` protected


package.json
```diff
@@ -23,8 +23,8 @@
     "bem": "~0.7.2",
     "mocha": "~1.9.0",
     "jshint-groups": "0.6.0",
-    "jscs": "1.10.0",
-    "jscs-jsdoc": "0.4.3",
+    "jscs": "1.11.3",
```
qfox, line 26: 1.12 is buggy?

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+                getEntityCls);
+
+        return instanceStorage[params.key] ||
+            (instanceStorage[params.key] = new EventManager(
+                params,
+                function(fn, fnId) {
+                    return function(e, data, flags, originalEvent) {
+                        if(!flags[fnId]) {
+                            originalEvent.data = e.data;
+                            originalEvent.bemTarget = $(e.currentTarget).bem(params.entityCls);
```
qfox: Oh.

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+
+    buildBemEventManagerParams = function(ctx, defCtx, defSelector, defCls, getEntityCls) {
+        var res = {
+                entityCls : null,
+                ctx : defCtx,
+                selector : defSelector,
+                key : ''
+            };
+
+        if(ctx) {
```
qfox: Maybe early returns?

``` js
if(!ctx) {
    res.entityCls = defCls;
    return res;
}
// etc...
```

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+         * @param {Object} [data] Additional data that the handler gets as e.data
+         * @param {Function} fn Handler
+         * @returns {EventManager} this
+         */
+        on : function(e, data, fn) {
+            if(typeof e === 'object') {
+                objects.each(e, function(fn, e) {
+                    this.on(e, data, fn);
+                }, this);
+            } else {
```
qfox: I'm not sure why you avoid early returns but code would look much simpler with them.

If we `return this` if that if we could move out `else` clause and "untab" the rest of function.

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+         * @param {String} [e] Event type
+         * @param {Function} [fn] Handler
+         * @returns {EventManager} this
+         */
+        un : function(e, fn) {
+            if(typeof e === 'object') {
+                objects.each(e, function(fn, e) {
+                    this.un(e, fn);
+                }, this);
+            } else {
```
qfox: Again. This method would look MUCH easier with early returns.

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+                var argsLen = arguments.length;
+                if(!argsLen) {
```
qfox: Also, `if(!condition) { /* some */ } else { /* other */ }` could be also rewritten as `if(condition) { /* other */ } else { /* some */ }`.

Or (better) with early returns.

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
qfox: Can we avoid jquery usage in event managers? ;-(

Can we use `i-bem-dom` itself for typecasting instead of 3rd-party (even if global and popular) libraries?

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_dom.js
```diff
@@ -0,0 +1,218 @@
...
+
+var undef,
+    winNode = window,
+    docNode = document,
+    winId = identify(winNode),
+    docId = identify(docNode),
+
+    eventStorage = {},
+
+    buildDomEventManagerParams = function(ctx, defCtx, defSelector, defCls, getEntityCls) {
```
qfox: Feels like a big part of it (and of buildBemEventManagerParams) can be moved out to some function.

common.blocks/i-bem-dom/__event-manager/_type/i-bem-dom__event-manager_type_bem.js
```diff
@@ -0,0 +1,231 @@
...
+                if(typeOfCtx === 'string') { // elem name
+                    blockName = defCls._blockName;
+                    elemName = ctx;
+                } else if(typeOfCtx === 'object') { // bem entity with optional mod val
+                    blockName = ctx.block?
+                        ctx.block.getName() :
+                        defCls._blockName;
+                    elemName = typeof ctx.elem === 'function'?
+                        ctx.elem.getName() :
+                        ctx.elem;
```
qfox: Complexity over 9k.

Maybe like that? https://gist.github.com/zxqfox/7cbb43addf13473ddbba

qfox:
Great work here, folks!

common.blocks/i-bem-dom/i-bem-dom.spec.js
```diff
@@ -1030,318 +1028,7 @@ describe('i-bem-dom', function() {
...
-
-            spyInit.called.should.be.true;
-            spyHandler.called.should.be.true;
-
-            BEMDOM.destruct(rootNode);
-            delete BEM.entities['block1'];
-            delete BEM.entities['block2'];
-        });
-    });
+    // TODO: add specs for live init
```
narqo: @dfilatov @veged lets do it here as well

narqo:
Two random specs failed with "Mocha init timeout", because of the CORS problems described in bem/bem-components#1528.

/cc @blond why travis status is green here?

blond:
@narqo update dependency does not help?

narqo:
Ok, the problem with status will be discussed in #1033. Everything looks fine here.

PR #979, opened 2015-04-15
Attribute escaping optimization #961

sbmaxx:
Fixes "Attribute escaping optimization #961".

All tests are fine.


narqo:
Could you rebase it after #981

sbmaxx:
@narqo done

also, should i port https://github.com/bem/bem-core/issues/982 from `v3`?

narqo:
> also, should i port #982 from v3?

~~@aristov already did it with #742~~
Oops, missread the original question. I think yes, we could port #742 to v2

narqo:
:fire:

PR #990, opened 2015-04-27
tick: example fixed in the description (close #978)

### LINKED ISSUE #978: tick: Fix examples to be consistent with idle
As we have fixed examples in `idle` block in #942 we need to do the same stuff with `tick`

Murdalay:


narqo:
:fire:

PR #1014, opened 2015-05-15
fix: example of inheritance

belozer:


common.docs/i-bem-js/i-bem-js.ru.md
```diff
@@ -494,7 +494,7 @@ provide(BEMDOM.decl(this.name, {}));
 
 modules.define('bblock', ['i-bem__dom', 'ablock'], function(provide, BEMDOM, ABlock) {
 
-provide(BEMDOM.decl({ block : this.name, baseBlock : ABlock ));
+provide(BEMDOM.decl({ block : this.name, baseBlock : ABlock }, {}));
```
narqo: There is nothing wrong in the original code. Everything but first argument is optional in `BemDom.decl`

belozer: This commit fix the second bracket
 { block : this.name, baseBlock : ABlock **}** <- this

tadatuta:
@belozyorcev thank you!

PR #1033, opened 2015-06-04
[WIP] Update npm devDependencies to mirror v2

narqo:


narqo:
@blond I've took all modules from v2, but it looks like status still doesn't work properly. Could you point me, what version of what packages should I update?

narqo:
@blond ping!

blond:
@narqo, I think that bug in `coa@1.0.1` (used in `enb-magic-platform@0.5.0`):
- [here](https://github.com/veged/coa/blob/master/lib/cmd.js#L240) `code` value is equal to `1`
- but [this code](https://github.com/veged/coa/blob/master/lib/cmd.js#L242-L245) is not executed

Also, when you try to return rejected promise the console displays: 

```
child_process: customFds option is deprecated, use stdio instead.
```

/cc @arikon @veged

blond:
A quick way to fix — use `magic run specs && magic run tmpl-specs` instead of `magic run specs tmpl-specs`.

narqo:
@blond could you make a separate PR into v2 branch with the fix?

blond:
@narqo #1040

blond:
In `coa` there is no bug.

This error somehow associated with the `enb-bem-tmpl-specs` package.

If after running `tmpl-specs` to call `process.exit(1);` then the console displays:

```
child_process: customFds option is deprecated, use stdio instead.
```

narqo:
Ok, #1040 looks fine, thanks. I'll port it to v3. Could you open a separate issue in `enb-bem-tmpl-specs` about fail positive results?

blond:
> Could you open a separate issue in enb-bem-tmpl-specs about fail positive results?

https://github.com/enb-bem/enb-bem-tmpl-specs/issues/88

PR #1035, opened 2015-06-05
identify: Get rid of onlyGet argument

### LINKED ISSUE #1028: identify: Get rid of onlyGet argument
As we don't use `onlyGet` argument anywhere, we can remake the API of the module to works like this `identify(...args)`. 

This will be help us to reduce copy-paste like `var fnId = identify(fn) + (fnCtx? identify(fnCtx) : '')`

narqo:
closes #1028

Plus, use new `identify` API.


narqo:
Ok, looks like everything is ok, except jquery loading from `yastatic.net`, which is described in bem/bem-components#1528.

PR #1057, opened 2015-06-13
page: Fix x-ua-compatible support

### LINKED ISSUE #794: page.bemhtml: x-ua-compatible is broken
New <meta> is created instead of updating the existed one when changing `x-ua-compatible` field to arbitrary string. 
head: { elem: 'head', 'x-ua-compatible': false } just do nothing.

Is that a bug or something wrong with BEMJSON examples?

tadatuta:
closes #794 
backport from v2


narqo:
:ok:

PR #1063, opened 2015-06-16
i-bem: fix jsDoc about getMod method

### LINKED ISSUE #1027: i-bem: fix getMod jsDoc
[As described in jsDoc](https://github.com/bem/bem-core/blob/v2/common.blocks/i-bem/i-bem.vanilla.js#L299) `getMod` returns strings only. But it returns boolean `true` for existing boolean modifier. It seems that we should fix jsDoc by adding `Boolean` type to the list of returned value types.
Yet another unclear thing is why `getMod` [returns an empty string](https://github.com/bem/bem-core/blob/v3/common.blocks/i-bem/i-bem.spec.js#L130) for missing (`false`) modifier, when boolean `false` is expected? Should we fix this behaviour too, or is there any reason why it works so?

aristov:
closes #1027


aristov:
Need port to v3.

PR #1065, opened 2015-06-18
Add npm-shrinkwrap.json to lock npm dependencies

narqo:
npm-shrinkwrap should prevent accident breaks after npm dependencies update (current `xjst` version, for example)

/jfi @tadatuta

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,6480 @@
+{
+  "name": "bem-core",
+  "version": "2.5.1",
+  "dependencies": {
+    "bem": {
+      "version": "0.9.0",
+      "from": "bem@>=0.9.0 <0.10.0",
+      "dependencies": {
+        "coa": {
+          "version": "0.4.1",
+          "from": "coa@0.4.1"
+        },
+        "q": {
+          "version": "0.9.7",
+          "from": "q@0.9.7"
+        },
+        "qq": {
+          "version": "0.3.5",
+          "from": "qq@0.3.5",
+          "dependencies": {
+            "q": {
+              "version": "0.8.4",
+              "from": "q@0.8.4"
+            }
+          }
+        },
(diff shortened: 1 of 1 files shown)


narqo:
Everything works with `xjst@1.5.0` locked

tadatuta:
Not sure about it.
1. We agreed with @indutny to use strong deps in `bem-xjst` on `xjst`
2. It's better to find regression in our libs before our users
3. Shrinkwrap won't help our users anyway but will make us to regenerate it each time we need to update any npm dependency.

narqo:
But at the same time, I don't understand, how this strict dependency helps us? 

1) We are not depend from `bem-xjst` in bem-components, and currently we depend from `bem-xjst@0.9` and `enb-bemxjst@1.3.4` in bem-core. What (and by who) should be done, so our PRs in bem-core  start working? 
2) If all dependencies in bem-xjst and other bem-enb-blahblah libraries will be strict what should be done (and by who) to test that the new version of `xjst` doesn't break bem-core?

Can't we use project-stub with semver dependencies or any other repo for such smoke testing?

tadatuta:
1. We have dependency on `bem-xjst` which had soft dependency on `xjst`. As it was replaced with strong dependency there's no change to get such breaking update any more (after update on new version of `bem-xjst` in `bem-core`).
2. Actually there are tests in `bem-xjst` and when broken `xjst` was released they failed. But that hadn't prevented us from getting that `xjst` version with soft deps. Then we have tests in `enb-bemxjst`. So with strong deps we will get an update just after all of them will pass.

narqo:
I still don't understand how does it all match to the phrase

>  It's better to find regression in our libs before our users

How could such regressions been found with bem-core if all our dependencies are strict?

narqo:
In any case I don't care much, my PR has been build finally with `xjst@1.5.3` :)

PR #1077, opened 2015-06-26
i-bem__dom: Fix deps

tadatuta:
#1076

Related https://github.com/enb-bem/enb-bem-techs/issues/158


narqo:
Let's add comment with a link to related bug in enb, so we can remove it after it will be fixed.

PR #1092, opened 2015-07-13
i-bem__dom: setMod doesn't add css class if blocks on the same DOM node have overlapping end parts in their names

### LINKED ISSUE #1090: i-bem__dom: setMod doesn't add css class if blocks on the same DOM node have overlapping end parts in their names
``` js
/*
<div class="modal personal-modal i-bem"></div> 
*/

BemDom.decl('personal-modal', {
  onSetMod : {
    'js' : {
      'inited' : function() {
        var modal = this.findBlockOn('modal');
        modal.hasMod('js', 'inited');   // true
        modal.domElem.hasClass('modal_js_inited');  // false  (Oops!, no "modal_js_inited" class here)
      }
    }
});
```

I believe this [`className.indexOf(classPrefix) > -1`](https://github.com/bem/bem-core/blob/ed70380c33febe093fa0407e1adb5695c5412449/common.blocks/i-bem/__dom/i-bem__dom.js#L701) is the problem.

/cc @4ok

narqo:
fix #1090

/cc @dfilatov

### DIFF
--- a/common.blocks/i-bem/__dom/i-bem__dom.js
+++ b/common.blocks/i-bem/__dom/i-bem__dom.js
@@ -698,7 +698,7 @@ DOM = BEM.decl('i-bem__dom',/** @lends BEMDOM.prototype */{
 
                 (oldModVal === true?
                     classRE.test(className) :
-                    className.indexOf(classPrefix + MOD_DELIM) > -1)?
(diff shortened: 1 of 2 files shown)


dfilatov:
:+1:

PR #1103, opened 2015-07-21
i-bem__dom: implement detach() method

narqo:
closes #1102

/cc @dfilatov

PR #1104, opened 2015-07-21
i-bem-dom: implement detach() method

### LINKED ISSUE #1102: i-bem__dom: implement `detach` method

narqo:
port #1103 into v3
close #1102

### DIFF
--- a/common.blocks/i-bem-dom/i-bem-dom.js
+++ b/common.blocks/i-bem-dom/i-bem-dom.js
@@ -883,11 +883,12 @@ bemDom = /** @exports */{
     },
 
     /**
-     * Destroys blocks on a fragment of the DOM tree
      * @param {jQuery} ctx Root DOM node
      * @param {Boolean} [excludeSelf=false] Exclude the main domElem
+     * @param {Boolean} [destructDom=false] Remove DOM node during destruction
+     * @private
      */
-    destruct : function(ctx, excludeSelf) {
+    _destruct : function(ctx, excludeSelf, destructDom) {
         var _ctx;
         storeDomNodeParents(_ctx = excludeSelf? ctx.children() : ctx);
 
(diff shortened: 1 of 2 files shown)


common.blocks/i-bem-dom/i-bem-dom.js
```diff
@@ -905,9 +914,8 @@ bemDom = /** @exports */{
         });
 
         // NOTE: it was moved here as jquery events aren't triggered on detached DOM elements
-        excludeSelf?
-            ctx.empty() :
-            ctx.remove();
+        _destructDom &&
+            (excludeSelf? ctx.empty() : ctx.remove());
```
narqo, line 910: @dfilatov do you remember why can't we do similar things in `v2` in the same way?

common.blocks/i-bem-dom/i-bem-dom.js
```diff
@@ -883,11 +883,20 @@ bemDom = /** @exports */{
...
     destruct : function(ctx, excludeSelf) {
+        this.detach(ctx, excludeSelf, true);
+    },
+
+    /**
+     * Detaches blocks on a fragment of the DOM tree
+     * @param {jQuery} ctx Root DOM node
+     * @param {Boolean} [excludeSelf=false] Exclude the main domElem
+     */
+    detach : function(ctx, excludeSelf, _destructDom) {
```
dfilatov: I would prefer not to expose `_destructDom`. Why not as in v2?

common.blocks/i-bem-dom/i-bem-dom.js
dfilatov, line 910: It was related to a new event system. But I can't remember how exactly.

common.blocks/i-bem-dom/i-bem-dom.js
narqo: I thought it would be easier, but guess it's not. Remade

dfilatov:
:ok:

PR #1120, opened 2015-08-12
i-bem.js-common: anchors corrected

innabelaya:


innabelaya:
cc @tadatuta @aristov

innabelaya:
#1119

PR #1154, opened 2015-08-24
ENB: removed underscore for built files

### LINKED ISSUE #1128: Switch to min.{css,js} instead of leading underscore for built files
...to be more like others and get rid of `.nojekyll` on `gh-pages`

blond:
Resolved #1128

### DIFF
--- a/.enb/helpers/page.js
+++ b/.enb/helpers/page.js
@@ -136,21 +136,21 @@ module.exports = function(node, opts) {
     }
 
     node.addTargets([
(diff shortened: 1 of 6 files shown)


sipayRT:
@blond what about examples? For example, https://github.com/bem/bem-core/blob/v2/common.blocks/page/page.examples/10-simple.bemjson.js

blond:
@sipayRT thanks, it will be fixed

PR #1224, opened 2015-10-02
i-bem: store locals on reapply

indutny:
Fix: https://github.com/bem/bem-xjst/issues/97

### DIFF
--- a/common.blocks/i-bem/i-bem.bemhtml
+++ b/common.blocks/i-bem/i-bem.bemhtml
@@ -250,11 +250,41 @@ BEMContext.prototype.generateId = function generateId() {
 var oldApply = exports.apply;
 exports.apply = BEMContext.apply = function BEMContext_apply(context) {
     var ctx = new BEMContext(context || this, oldApply);
+
     ctx.apply();
+
     return ctx._str;
 };
 
-BEMContext.prototype.reapply = BEMContext.apply;
+BEMContext.prototype.reapply = function reapply() {
+    var mode = typeof $$mode !== 'undefined' && $$mode,
+        block = typeof $$block !== 'undefined' && $$block,
+        elem = typeof $$elem !== 'undefined' && $$elem,
+        elemMods = typeof $$elemMods !== 'undefined' && $$elemMods,
+        mods = typeof $$mods !== 'undefined' && $$mods,
+
+        classicMode = this._mode,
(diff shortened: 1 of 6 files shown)


arikon:
@veged @tadatuta Could you make review plz?

arikon:
@mishaberezin This should be ported to `islands` to be fixed there

veged:
@indutny LGTM about standard fields, but what about custom fields in context?

tadatuta:
Added custom field to test, everything works fine. Let's merge!

arikon:
@tadatuta I think, it should be better to squash commits

tadatuta:
@arikon squashed, rebased

tadatuta:
@mishaberezin have you ported it to Islands?

mishaberezin:
Nope. Going to port on next week.

common.blocks/i-bem/i-bem.bemhtml
```diff
@@ -250,11 +250,41 @@ BEMContext.prototype.generateId = function generateId() {
...
     var ctx = new BEMContext(context || this, oldApply);
+
     ctx.apply();
+
     return ctx._str;
 };
 
-BEMContext.prototype.reapply = BEMContext.apply;
+BEMContext.prototype.reapply = function reapply() {
+    var mode = typeof $$mode !== 'undefined' && $$mode,
```
qfox, line 260: @indutny @tadatuta Why globals? What if I call reapply twice()? Shouldn't it be a some kind of states stack?

indutny, line 260: @zxqfox this was introduced for 20% speed boost. `BEMContext` is way too polymorphic, and access these properties was very slow.

qfox, line 260: We can define them all (`$$*`) as variables in the parent scope (inside `oninit`). I don't see why they should be globals and how it (making them globals) gives 20% speed boost ;-(

indutny, line 260: I forgot to say, they are not globals. They are in parent scope of `oninit()`, which is non-global.

qfox, line 260: Ah, thanks for noticing. Now it's much clearer.

PR #1229, opened 2015-10-13
page__css.bemhtml: Support new bem-xjst versions

### LINKED ISSUE #1228: page__css.bemhtml: Support new bem-xjst versions

tadatuta:
Closes #1228


blond:
:+1:

PR #1239, opened 2015-11-13
Update linters @v3

qfox:
Ref #1235 
Closes #1043

PR #1284, opened 2016-04-01
Update linters @v4

qfox:
Rebased https://github.com/bem/bem-core/pull/1239

### DIFF
--- a/.githooks/pre-commit/lint
+++ b/.githooks/pre-commit/lint
@@ -21,7 +21,7 @@ trap cleanup EXIT SIGINT SIGHUP
 git diff --no-prefix > "$PATCH_FILE"
 git checkout -- .
 
-git_cached_files=$(git diff --cached --name-only --diff-filter=ACMR | xargs echo)
+git_cached_files=$(git diff --cached --name-only --diff-filter=ACMR | grep -e \.js$ -e \.bemhtml$ -e \.bemtree$ | xargs echo)
 if [ "$git_cached_files" ]; then
     $NPM_BIN/jshint-groups $git_cached_files || exit 1
     $NPM_BIN/jscs $git_cached_files || exit 1
--- a/common.blocks/i-bem/__collection/i-bem__collection.js
+++ b/common.blocks/i-bem/__collection/i-bem__collection.js
@@ -164,7 +164,7 @@ var BemCollection = inherit(/** @lends BemCollection.prototype */{
     /**
      * Returns a new collection comprised of collection on which it is called joined with
      * the collection(s) and/or array(s) and/or entity(es) provided as arguments.
-     * @param {Collection|Array|BemEntity} [args...]
+     * @param {?...(Collection|Array|BemEntity)} args
      * @returns {Collection}
      */
     concat : function() {
@@ -199,8 +199,6 @@ var BemCollection = inherit(/** @lends BemCollection.prototype */{
     }
 });
 
-var arrayConcat = Array.prototype.concat;
-
 function buildForEachEntityMethodProxyFn(methodName) {
     return function() {
         var args = arguments;
@@ -226,6 +224,8 @@ function buildComplexProxyFn(arrayMethodName, entityMethodName) {
         });
     };
 }
+
+var arrayConcat = Array.prototype.concat;
 
 provide(BemCollection);
 
--- a/common.blocks/i-bem/i-bem.vanilla.js
+++ b/common.blocks/i-bem/i-bem.vanilla.js
@@ -182,6 +182,8 @@ function declEntity(baseCls, entityName, base, props, staticProps) {
     return entityCls;
 }
 
+// jscs:disable requireMultipleVarDecl
+
 /**
  * @class BemEntity
  * @description Base block for creating BEM blocks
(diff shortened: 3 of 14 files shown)


common.blocks/i-bem/__collection/i-bem__collection.js
```diff
@@ -179,7 +179,7 @@ var BemCollection = inherit(/** @lends BemCollection.prototype */{
                 arg instanceof BemCollection?  arg._entities : arg);
         }
 
-        return new BemCollection(arrayConcat.apply(this._entities, argsForConcat));
+        return new BemCollection(Array.prototype.concat.apply(this._entities, argsForConcat));
```
narqo: Why have you decided to do it? It doesn't look like a codestyle fix.

common.blocks/i-bem/__collection/i-bem__collection.js
```diff
@@ -164,7 +164,7 @@ var BemCollection = inherit(/** @lends BemCollection.prototype */{
     /**
      * Returns a new collection comprised of collection on which it is called joined with
      * the collection(s) and/or array(s) and/or entity(es) provided as arguments.
-     * @param {Collection|Array|BemEntity} [args...]
+     * @param {?...(Collection|Array|BemEntity)} args
```
narqo, line 167: Could you point to specification of JSDoc with this syntax. I'm not sure `jsd` supports this syntax.

common.blocks/i-bem/__collection/i-bem__collection.js
qfox: This variable used just once, but placed in separated var statement that causes jscs error.

common.blocks/i-bem/i-bem.vanilla.js
```diff
@@ -182,6 +182,8 @@ function declEntity(baseCls, entityName, base, props, staticProps) {
     return entityCls;
 }
 
+// jscs:disable requireMultipleVarDecl
```
narqo, line 185: Why do we have to disable `requireMultipleVarDecl`. We don't use it in our lint rules, do we?

common.blocks/identify/identify.vanilla.js
```diff
@@ -27,7 +27,7 @@ provide(
     /**
      * Makes unique ID
      * @exports
-     * @param {Object} [obj...] Object that needs to be identified
+     * @param {?...Object} obj Object that needs to be identified
```
narqo, line 30: Same question about JSDoc syntax and jsd here.

common.blocks/i-bem/__collection/i-bem__collection.js
qfox, line 167: http://usejsdoc.org/tags-type.html

Look for `{?`, `{(`.

common.blocks/i-bem/i-bem.vanilla.js
qfox, line 185: https://github.com/bem/jscs-bem/blob/master/presets/bem.json#L4-L5
https://github.com/bem/jscs-bem/blob/master/test/presets/bem.js#L22-L58

common.blocks/i-bem/__collection/i-bem__collection.js
qfox: :up:

veged:
@zxqfox can we merge it?

qfox:
@veged Guess, yeah!

.githooks/pre-commit/lint
```diff
@@ -21,8 +21,8 @@ trap cleanup EXIT SIGINT SIGHUP
 git diff --no-prefix > "$PATCH_FILE"
 git checkout -- .
 
-git_cached_files=$(git diff --cached --name-only --diff-filter=ACMR | xargs echo)
+git_cached_files=$(git diff --cached --name-only --diff-filter=ACMR | grep \.js$ -e \.bemhtml$ -e \.bemtree$ | xargs echo)
```
narqo: @zxqfox looks like the `-e` switch has been missed before `\.js$`. Current snippet leads to

> grep: \.js$: No such file or directory

in case of no JS files was modified.

qfox: Yeah, fixed.

narqo:
Looks like it fixes #1043 as well, isn't it?

common.blocks/i-bem/__collection/i-bem__collection.js
narqo, line 167: BTW, `{?Collection}` means that parameter can be a `Collection` or **null**, e.g. "Nullable type". Optional parameter must be declared as `{Collection} [args]` (or `{Collection=}`, which is weird).

qfox, line 167: Guess, **nullable** is the same as **undefinedable**, so there is no real difference. Isn't it?

narqo, line 167: Not exactly. The difference is in `arguments.length`. 

For example, the first argument of most nodejs' callback functions is an error object, which is the nullable argument (it can be `null`, but the argument must be passed to the function). "Optional argument" means that it can be omitted, but if it presents, it must be of a proper type.

```
› [1].concat()
‹ [1]
› [1].concat(null)
‹ [1, null]
› [1].concat(undefined)
‹ [1, undefined]
```

qfox, line 167: Well, I know how JS works. But there is no difference for documentation. ;-) Btw, `undefined == null`.

narqo, line 167: What I've tried to explain is that cases below are valid according to `{?Collection}` notation but they won't work.

``` js
collection.concat(null).findChildBlock(···); // will throw
collection.concat(undefined).findChildBlock(···); // will throw
```

At the same time, this case is only valid for `{Collection} [arg]` and will work as expected by API

``` js
collection.concat().findChildBlock(···); // will work
```

PR #1353, opened 2016-05-11
i-bem-dom: Make result of `_elem*` to be union of `findChildElem*` and `findMixedElem*`

### LINKED ISSUE #1346: i-bem-dom: Make result of `_elem*` to be union of `findChildElem*` and `findMixedElem*`
As minor consequence we have to drop cache refresh optimisation (from #583).

veged:
closes #1346


veged:
need specs

PR #1390, opened 2016-07-19
Prevent block re-initialization after it had been destructed

### LINKED ISSUE #1383: Prevent block re-initialization after it had been destructed
E.g. here https://github.com/bem/bem-components/blob/v3/common.blocks/select/select.js#L154-L155 `button` will be already destructed.

narqo:
fix #1383

### DIFF
--- a/common.blocks/i-bem-dom/i-bem-dom.spec.js
+++ b/common.blocks/i-bem-dom/i-bem-dom.spec.js
@@ -1233,6 +1233,30 @@ describe('i-bem-dom', function() {
             bemDom.destruct(blockNode);
             spy.should.have.been.called;
         });
+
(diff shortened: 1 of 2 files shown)


veged:
sure!

common.blocks/i-bem-dom/i-bem-dom.spec.js
```diff
@@ -1233,6 +1233,30 @@ describe('i-bem-dom', function() {
             bemDom.destruct(blockNode);
             spy.should.have.been.called;
         });
+
+        // see https://github.com/bem/bem-core/issues/1383
+        it.only('should not re-initialize destructing entities during destruct', function() {
```
narqo: only!

PR #1404, opened 2016-08-31
Improve use of buildClassName fn

Yeti-or:
Just a little improvement.
`buildClassName` has a lot of checks before it finally gets to case with only block variable.
It seems that we can omit `buildClassName` for such cases.

``` js
    buildClassName : function(block, elem, modName, modVal) {
        if(isSimple(modName)) {
            if(!isSimple(modVal)) {
                modVal = modName;
                modName = elem;
                elem = undef;
            }
        } else if(typeof modName !== 'undefined') {
            modName = undef;
        } else if(elem && typeof elem !== 'string') {
            elem = undef;
        }

        if(!(elem || modName)) { // optimization for simple case
            return block;
        }
        ...
```


Yeti-or:
@tadatuta you don't like such change?

tadatuta:
@Yeti-or thank you!

PR #1407, opened 2016-09-05
Fix possible double initialization (close #1405)

### LINKED ISSUE #1405: lazyInit on _js_inited works wrong
Actually I don't understand what's going on but in `checkbox` code based on `bem-core@v3` there is a line https://github.com/bem/bem-components/blob/v3/common.blocks/radio/_type/radio_type_button.js#L37 which I ported to `bem-core@v4` as https://github.com/bem/bem-components/blob/v5/common.blocks/radio/_type/radio_type_button.js#L39

As a result in `radio-group` the `checkbox` which was checked first can never be checked again (first time it works just fine as well as any other checkboxes of the group). And for some reason it gets `_focused-hard` even being clicked with a cursor.

All the blocks are properly inited though and any other events like `this._events(Button).on({ modName : 'hovered', modVal : true }, Functions.noop);` work fine.

dfilatov:


dfilatov:
🚧 WIP

dfilatov:
cc @narqo @veged

Issue #1406, opened 2016-09-05
BEMHTML: Support escaped mode


tadatuta:
In `.enb/make.js` for BEMHTML:

``` js
{
    engineOptions: { escapeContent: true }
}
```

It fails at least for doctype in `page` template.

qfox:
You should check for html-entities (and replace them with utf-8 characters in common), BEMHTML.reapply & BH.apply & manually written scripts in ua block (replace with `{tag: false, html: ...}`). Maybe something more.

PR #1409, opened 2016-09-05
BEMHTML: Support escaped mode #1406

tadatuta:


.enb/tasks/tmpl-specs.js
```diff
@@ -65,6 +79,16 @@ function configure(helper, platform) {
                     },
                     engineOptions : { elemJsInstances : true }
                 }
+            },
+            bemhtmlEscaped : {
+                tech : 'enb-bemxjst/techs/bemhtml',
+                options : {
+                    sourceSuffixes : ['bemhtml', 'bemhtml.js'],
+                    engineOptions : { escapeContent : true },
```
qfox: А почему без `elemJsInstances : true`? Или зачем они выше?

package.json
```diff
@@ -31,7 +31,7 @@
     "enb-bem-specs": "^0.9.0",
     "enb-bem-techs": "^2.1.0",
     "enb-bem-tmpl-specs": "^1.1.0",
-    "enb-bemxjst": "^6.3.0",
+    "enb-bemxjst": "^6.5.3",
     "enb-bemxjst-i18n": "1.0.0-beta3",
```
qfox, line 35: It uses enb-bemxjst@2.1.0
Why we still using it?

.enb/tasks/tmpl-specs.js
tadatuta: добавил, спасибо

package.json
tadatuta: discussed in chat (package should be updated but we don't have time for it now)

PR #1412, opened 2016-09-06
Use jQuery 3.1

### LINKED ISSUE #1371: Use jQuery 3.1

tadatuta:
Closes #1371

### DIFF
--- a/common.blocks/i-bem-dom/__events/_type/i-bem-dom__events_type_dom.spec.js
+++ b/common.blocks/i-bem-dom/__events/_type/i-bem-dom__events_type_dom.spec.js
@@ -651,7 +651,7 @@ describe('DOM events', function() {
                                     .on('click', spy1)
                                     .on('click', spy2);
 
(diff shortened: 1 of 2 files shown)


narqo (changes requested):
What about this code for old IEs https://github.com/bem/bem-core/blob/331d49c80a01157e1126b6bfad77a740eded8f78/desktop.blocks/jquery/__config/jquery__config.js#L16 

It looks like it'd break.

tadatuta:
@narqo what exactly should break?

narqo:
You've added changes:

``` diff
-                block1._domEvents(rootNode.find('div').andSelf()).un('dblclick');
+                block1._domEvents(rootNode.find('div').addBack()).un('dblclick');
```

Does jQuery 1.10 support this API?

tadatuta:
I think it doesn't matter as it's in `spec.js` which we run only in phantom anyways.

PR #1441, opened 2016-10-24
document.uniqueID always returns unique id

### LINKED ISSUE #1440: "click" event handler accumulate after activation in IE11
Code:

``` js
this._events(Button).on('click', function () {

}, this);
```

After `n`'th click we have `n+1` click handlers.

vsesh:
Fix #1440

### DIFF
--- a/common.blocks/identify/identify.vanilla.js
+++ b/common.blocks/identify/identify.vanilla.js
@@ -6,15 +6,19 @@ modules.define('identify', function(provide) {
 
 var counter = 0,
     expando = '__' + (+new Date),
+    global = this.global,
     get = function() {
         return 'uniq' + (++counter);
     },
     identify = function(obj) {
         if((typeof obj === 'object' && obj !== null) || typeof obj === 'function') {
-            var key = 'uniqueID' in obj && obj !== document ?
-                'uniqueID' :
-                expando; // Use when possible native uniqueID for elements in IE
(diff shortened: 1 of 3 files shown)


veged (changes requested):
this block supposed to use in "vanilla.js" runtime that lacks `document` object — so let's do the check in more safe cross-engine way

common.blocks/identify/identify.vanilla.js
```diff
@@ -11,7 +11,7 @@ var counter = 0,
     },
     identify = function(obj) {
         if((typeof obj === 'object' && obj !== null) || typeof obj === 'function') {
-            var key = 'uniqueID' in obj?
+            var key = 'uniqueID' in obj && obj !== document ?
```
dfilatov: I think there's a better way:

``` js
var key;
if('uniqueID' in obj) {
    obj === global.document && (obj = obj.documentElement); // global should point to this.global inside module closure
    key = 'uniqueID';
} else {
    key = expando;
}
```

dfilatov: As a bonus, you won't have to check browser environment.

veged: LGTM

vsesh: ok

PR #1469, opened 2017-01-10
FindChildBlock throws an Error if Block given as string

skad0:
Referred to [this](https://github.com/bem/bem-core/issues/1464)

@tadatuta

### DIFF
--- a/common.blocks/i-bem-dom/i-bem-dom.js
+++ b/common.blocks/i-bem-dom/i-bem-dom.js
@@ -312,6 +312,20 @@ function getJqueryCollection(html) {
 }
 
 /**
+ * Validates block to be class or specified description
+ * @param {*} Block Any argument passed to find*Block as Block
+ * @throws {Error} Will throw an error if the Block argument isn't correct
+ */
+function validateBlockParam(Block) {
+    if(
+        typeof Block === 'string' ||
+        typeof Block === 'object' && typeof Block.block === 'string'
+    ) {
+        throw new Error('Block must be a class or description (block, modName, modVal) of the block to find');
+    }
+}
+
+/**
  * @class BemDomEntity
  * @description Base mix for BEM entities that have DOM representation
  */
@@ -442,7 +456,8 @@ var BemDomEntity = inherit(/** @lends BemDomEntity.prototype */{
      * @returns {Block}
      */
     findChildBlock : function(Block) {
-        // TODO: throw if Block passed as a string
+        validateBlockParam(Block);
+
         return this._findEntities('find', Block, true);
     },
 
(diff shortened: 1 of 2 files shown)


tadatuta:
@skad0 should also throw on `{ block: 'string' }`

common.blocks/i-bem-dom/i-bem-dom.js
```diff
@@ -442,7 +442,10 @@ var BemDomEntity = inherit(/** @lends BemDomEntity.prototype */{
      * @returns {Block}
      */
     findChildBlock : function(Block) {
-        // TODO: throw if Block passed as a string
+        if(typeof Block === 'string') {
+            throw new Error('Block can\'t be a String');
```
narqo: The message which describes what type `Block` is allowed to be, might be more useful in case of error. Consider "Block must be a class or an instance", as an example.

skad0:
:up:

tadatuta:
cc @dfilatov

dfilatov:
* What about this check in others `find*Block` methods?
  * I think this check should be extracted to a separate function, something like `validateBlockDesc` and used in each `find*Block` method.

skad0:
:up:

dfilatov:
lgtm

common.blocks/i-bem-dom/i-bem-dom.js
```diff
@@ -312,6 +312,20 @@ function getJqueryCollection(html) {
...
+ * @param {*} Block Any argument passed to find*Block as Block
+ * @throws {Error} Will throw an error if the Block argument isn't correct
+ */
+function validateBlockParam(Block) {
+    if(
+        typeof Block === 'string' ||
+        typeof Block === 'object' && typeof Block.block === 'string'
+    ) {
+        throw new Error('Block must be a class or description (block, modName, modVal) of the block to find');
+    }
```
birhoff, line 325: - i still can pass primitives numbers, boolean, undefined 
 - any function

error message of this validation and validation description don't fit with code.

common.blocks/i-bem-dom/i-bem-dom.js
```diff
@@ -442,7 +456,8 @@ var BemDomEntity = inherit(/** @lends BemDomEntity.prototype */{
      * @returns {Block}
      */
     findChildBlock : function(Block) {
-        // TODO: throw if Block passed as a string
+        validateBlockParam(Block);
+
         return this._findEntities('find', Block, true);
```
birhoff, line 461: Why we don't do this check at `_findEntities `?

birhoff (changes requested):

common.blocks/i-bem-dom/i-bem-dom.js
tadatuta: 'cos `_findEntities` is also used for elements passed as strings

PR #1489, opened 2017-03-07
i-bem-dom__collection: Fix concat() method

### LINKED ISSUE #1488: i-bem-dom__collection concat is broken
`i-bem-dom__collection` inherits `i-bem__collection` and does not override `concat` method so it returns an instance of `i-bem__collection` (see https://github.com/bem/bem-core/blob/v4/common.blocks/i-bem/__collection/i-bem__collection.js#L182).

tadatuta:
Closes #1488

### DIFF
--- a/common.blocks/i-bem/__collection/i-bem__collection.js
+++ b/common.blocks/i-bem/__collection/i-bem__collection.js
@@ -179,7 +179,7 @@ var BemCollection = inherit(/** @lends BemCollection.prototype */{
                 arg instanceof BemCollection?  arg._entities : arg);
         }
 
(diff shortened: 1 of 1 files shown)


dfilatov:
Looks like it could be done more general and without overriding, replace inside `i-bem__collection`:
```js
 return new BemCollection(arrayConcat.apply(this._entities, argsForConcat));
```
with
```js
 return new this.__self(arrayConcat.apply(this._entities, argsForConcat));
```

tadatuta:
@dfilatov, you are right, thank you!

fixed.

veged (approved):

PR #1523, opened 2017-08-21
i-bem: Get rid of `_delInitedMod`

### LINKED ISSUE #1512: i-bem: Get rid of `_delInitedMod`
It is not used for overrides.

tadatuta:
Closes #1512


tadatuta:
cc @veged

PR #1528, opened 2017-09-10
idle: Get rid of jQuery

tadatuta:


common.blocks/idle/idle.js
```diff
@@ -27,7 +27,9 @@ var IDLE_TIMEOUT = 3000,
             if(!this._isStarted) {
                 this._isStarted = true;
                 this._startTimer();
-                $(document).on(USER_EVENTS, $.proxy(this._onUserAction, this));
+                USER_EVENTS.forEach(function(event) {
+                    document.addEventListener(event, this._onUserAction.bind(this));
```
veged: this `bind` better to do once in `__constructor` — `this._onUserAction = this._onUserAction.bind(this)` — so you can use same function in `addEventListener`/`removeEventListener` (wondered how previous variant works)

tadatuta:
@veged ⬆️

Issue #1526, opened 2017-09-08
Unsubscribe from particular event won't work for subscription of several events


tadatuta:
```js
this._domEvents().on('click', cb);
this._domEvents().un('click', cb);
```
works fine but
```js
this._domEvents().on('click keypress', cb);
this._domEvents().un('click', cb);
```
does not work as handler is kept in hash where key depends on event name.

veged:
Fixed in v5 (commit 79bc773): on/un now split space-separated event strings into individual events, allowing proper unsubscription of single events from multi-event subscriptions.

PR #1529, opened 2017-09-10
Fix multiple events subscription

tadatuta:
This PR also addresses #1526 for bem-core version without jQuery.

### DIFF
--- a/common.blocks/i-bem-dom/__events/i-bem-dom__events.js
+++ b/common.blocks/i-bem-dom/__events/i-bem-dom__events.js
@@ -52,6 +52,14 @@ var undef,
          * @returns {EventManager} this
          */
         on : function(e, data, fn, _fnCtx, _isOnce) {
+            if(typeof e === 'string' && e.indexOf(' ') > -1) {
+                e.split(' ').forEach(function(event) {
+                    this.on(event, data, fn, _fnCtx, _isOnce);
+                }, this);
+
+                return this;
+            }
+
             var params = this._params,
                 event = this._eventBuilder(e, params, 'on');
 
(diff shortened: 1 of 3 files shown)


common.blocks/i-bem-dom/__events/i-bem-dom__events.js
```diff
@@ -52,38 +52,44 @@ var undef,
...
+                        });
+                    }
+                }.bind(this),
+                handler;
 
-                bindDomNodes.forEach(function(domNode) {
-                    domNode.addEventListener(event, handler, false);
-                });
-            }
+            typeof e === 'string'?
```
veged: it's better to do this in very first lines of this function — just check if `e` is a string and contains spaces and do `forEach` of `this.on` (we do similar https://github.com/bem/bem-core/blob/v4/common.blocks/i-bem-dom/__events/i-bem-dom__events.js#L226)

common.blocks/i-bem-dom/__events/i-bem-dom__events.js
```diff
@@ -115,26 +121,32 @@ var undef,
...
-                        domNode.removeEventListener(event, handler);
-                    });
-                }
+                            bindDomNodes.forEach(function(domNode) {
+                                domNode.removeEventListener(event, handler);
+                            });
+                        }
+                    }.bind(this);
+
+                typeof e === 'string'?
```
veged: same stuff like with `on` — we can make smaller diff

tadatuta:
@veged ⬆️

common.blocks/i-bem-dom/__events/i-bem-dom__events.js
```diff
@@ -52,6 +52,14 @@ var undef,
          * @returns {EventManager} this
          */
         on : function(e, data, fn, _fnCtx, _isOnce) {
+            if(typeof e === 'string' && e.indexOf(' ') > -1) {
+                e.split(' ').forEach(function(evnt) {
```
veged: `evEnt`

veged (approved):

PR #1550, opened 2018-01-20
Fix bemDom decls with mixins

### LINKED ISSUE #1549: Bug: bemDom.declBlock with mixins
Bug case

```js
const Mixin = bemDom.declMixin({ myProp : 'test' });
Block = bemDom.declBlock('block', [Mixin]);

Block.prototype.findChildBlock // undefined
```

but

```js
const Mixin = bemDom.declMixin({ myProp : 'test' });
Block = bemDom.declBlock('block', [bemDom.Block, Mixin]);

Block.prototype.findChildBlock // function
```

After add code (*This is the draft version of the solution. It remains to do right now.*)

```js
    /**
     * Declares DOM-based block and creates block class
     * @param {String|Function} blockName Block name or block class
     * @param {Function|Array[Function]} [base] base block + mixes
     * @param {Object} [props] Methods
     * @param {Object} [staticProps] Static methods
     * @returns {Function} Block class
     */
    declBlock : function(blockName, base, props, staticProps) {
        if(!base || (typeof base === 'object' && !Array.isArray(base))) {
            staticProps = props;
            props = base;
            base = typeof blockName === 'string'?
                entities[blockName] || Block :
                blockName;
        }

        // added code start
        base || (base = entities[blockName] || Block);

        Array.isArray(base) || (base = [base]);

        if(!base[0].__bemEntity) {
            base = base.slice();
            base.unshift(entities[blockName] || Block);
        }
         // added code end

        return bem.declBlock(blockName, base, props, staticProps);
    },

```

```js
Block = bemDom.declBlock('block', [Mixin]);
Block.prototype.findChildBlock // function
```

P.S.
A problem can also be with elements

belozer:
fix #1549

### DIFF
--- a/common.blocks/i-bem-dom/i-bem-dom.js
+++ b/common.blocks/i-bem-dom/i-bem-dom.js
@@ -322,6 +322,26 @@ function validateBlockParam(Block) {
 }
 
 /**
+ * Returns base entities for declaration
+ * @param {Function} baseCls block|elem class
+ * @param {String} entityName entityName
+ * @param {Function|Array[Function]} [base] base block|elem + mixes
+ * @returns {Array<Function>}
+ */
+function getEntityBase(baseCls, entityName, base) {
+    base || (base = entities[entityName] || baseCls);
+
+    Array.isArray(base) || (base = [base]);
+
+    if(!base[0].__bemEntity) {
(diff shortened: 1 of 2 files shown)


belozer:
@veged

PR #1558, opened 2018-03-05
Deps: enb-bemxjst fixed and updated

miripiruni:
1) `options.requires` as fallback for `options.engineOptions.requires`
2) global dependencies with dot notation (`'BEM.I18N'` for example)


miripiruni:
ping

PR #1573, opened 2018-05-16
Fix set boolean mod on elem-instance

### LINKED ISSUE #1572: v3: bug set mod without val on elem-instance
Test case

```js
it('should set element\'s modifier without val', function() {
    DOM.decl('block', {}, {});
    DOM.decl({ block : 'block', elem : 'elem' }, {}, {});

    var rootNode = $(BEMHTML.apply({
            block : 'block',
            js : true,
            content : {
                elem : 'elem',
                mix : { block : 'i-bem' },
                js : true
            }
        })),
        block = rootNode.bem('block'),
        elem = block.elemInstance('elem');

    elem.hasMod('mod').should.be.false;

    block.setMod(block.elem('elem'), 'mod');
    elem.hasMod('mod').should.be.true;

    DOM.destruct(rootNode);
    delete DOM.blocks['block'];
    delete DOM.blocks['block__elem'];
});
```

Later I will try to send a fix

belozer:
Fix #1572

### DIFF
--- a/common.blocks/i-bem/__dom/_elem-instances/i-bem__dom_elem-instances.js
+++ b/common.blocks/i-bem/__dom/_elem-instances/i-bem__dom_elem-instances.js
@@ -64,6 +64,10 @@ provide(BEMDOM.decl(/** @lends BEMDOM.prototype */{
     setMod : function(elem, modName, modVal) {
         var elemClass;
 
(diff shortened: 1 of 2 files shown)


tadatuta (approved):

tadatuta:
👍

